PR #76, opened 2025-02-27
test: support acceptance tests and sweepers

sosheskaz:
* Use provided API token for acceptance tests
* Implement sweepers and Makefile target to clean up unused resources
* Extract underlying client construction in provider into separate function for reuse.
* Create `testutil` package for some common functions of tests.


coreweave/networking/resource_vpc_test.go
```diff
@@ -17,8 +21,91 @@ import (
...
+				deleteResp, err := client.DeleteVPC(ctx, deleteReq)
+				if err != nil {
+					return fmt.Errorf("failed to delete VPC %s: %w", vpc.Name, err)
+				}
+				deletedVpc := deleteResp.Msg.Vpc
+
+				timeout := time.After(1 * time.Minute)
+				ticker := time.NewTicker(10 * time.Second)
+				defer ticker.Stop()
+				for {
```
soggycactus, line 79: This is probably worth abstracting out into a function in the testutil package, a lot of resources will be asynchronous during their CRUD lifecycle - I'm thinking something that is similar to testify's [assert.Eventually](https://github.com/stretchr/testify/blob/v1.10.0/assert/assertions.go#L1930)

internal/testutil/env.go
```diff
@@ -0,0 +1,23 @@
...
+//
+// Returns:
+//   - bool: True if the environment variable was not previously set, false otherwise.
+func SetEnvIfUnset(name, value string) bool {
+	_, found := os.LookupEnv(name)
+	if !found {
+		os.Setenv(name, value)
+	}
+	return !found
+}
```
soggycactus, line 23: I'd rather we make this more opinionated - I don't think there's any reason each test should have to pass in the name and value, since those will always be the same. Would be better to be able to invoke the function without any arguments

internal/testutil/api_deleter.go
```diff
@@ -0,0 +1,60 @@
...
+
+// WithIntervalRetry repeatedly executes the provided function `f` at the specified `interval`
+// until it returns true or the context `ctx` is done. It uses a ticker to manage the interval
+// timing. If the context is canceled or times out, the function will return without retrying further.
+//
+// Parameters:
+// - ctx: The context to control the lifecycle of the retries. If the context is done, the retries stop.
+// - interval: The duration to wait between each retry attempt.
+// - f: The function to be executed repeatedly. It should return true to stop further retries, or an error to denote success.
+func WithIntervalRetry(ctx context.Context, interval time.Duration, f func() (bool, error)) error {
```
soggycactus, line 18: I think we can probably merge this function into wait for delete - this way the caller can set an appropriate interval themselves, as the time it takes to delete something can depend on the type resource itself

If you anticipate us using this retry logic more, you can pull that out into a separate private function - but I think the end user should pass the interval as well. Very clever to have them also pass a generic connect request since deletions are represented by 404s

soggycactus (approved):

PR #80, opened 2025-03-06
feat: add coreweave_cks_cluster data source

sosheskaz:


coreweave/cks/data_cluster.go
```diff
@@ -0,0 +1,273 @@
...
+
+func NewClusterDataSource() datasource.DataSource {
+	return &ClusterDataSource{}
+}
+
+type ClusterDataSource struct {
+	client *coreweave.Client
+}
+
+type ClusterDataSourceModel struct {
```
soggycactus, line 31: Is there a reason we can't use the already existing ClusterResourceModel? That way we can cut down a lot of this duplication

sosheskaz: I recalled reading something that suggested that decoupling models between resources and data sources was the preferred method... but can't find it now. Combining the models would have the benefit of failing tests for the data source if the model is updated, but the schema is not, so I think you're right, that would be better. It's also an easily reversible decision. Will implement.

soggycactus (approved):

soggycactus (approved):

PR #106, opened 2025-06-02
fix: update retry policy to allow context.DeadlineExceeded errors

soggycactus:
Updates the retry policy to retry timeout errors, and also allows the end user to configure the timeout interval via `http_timeout`


idodod (commented):
Left some suggestions for improvement but LGTM. Just wondering what tests look like in this repo?

PR #125, opened 2025-09-22
deps: bump go to 1.24.7

sosheskaz:


idodod (approved):
I am guessing theer is a reason why we're not bumping to 1.25?

PR #143, opened 2025-10-10
ci: improve caching for snapshot build

sosheskaz:
This changes the behavior of the snapshot build to improve PR experience.

The build logic is extracted into a reusable workflow and called from PRs as well as from a main build, to hydrate build caches in advance of PRs being opened.

It no longer uses the inline cache from `actions/setup-go`. Instead, it self-manages its cache. This means it builds slightly slower in the worst case (cache miss), but dramatically faster after the first build succeeds after a gomod change.

In the example, it took 8m1s to perform an ex nihilo build, but subsequent builds completed in 50s.

This also adds `build` as a dependency to `generate`, to take better advantage of caching.

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -1,7 +1,7 @@
 on:
   push:
     branches:
-      - main # running from the default branch makes the cache available to PRs.
+      - main # running from the default branch makes the go build cache available to PRs.
 
 jobs:
   build:


.github/workflows/main-build.yml
```diff
@@ -0,0 +1,11 @@
+on:
+  push:
+    branches:
+      - main # running from the default branch makes the cache available to PRs.
```
pmowrer: nit: It's unclear what "the cache" is w/o context

pmowrer (dismissed):

pmowrer (approved):

PR #159, opened 2025-11-07
fix: use prototext to format debug payloads

sosheskaz:
Prototext is the preferred solution for human-readable output, and is expected to respect settings which may control human-readable output. This is desireable for our debug logs. For example, omitting sensitive fields _should_ someday be supported by prototext, but may not be supported by protojson, because it is expected that protojson should always be able to re-marshal its output.

This also moves the code to a separate file, to keep client.go focused on functional behaviors of the client, and move log handling and testing into cohesive, single-purpose files.

A few other changes:
* Rename "payload" to "message" for consistent terminology.
* Use a symbol for the message key.
* Test the contents of debugged payloads.


coreweave/loginterceptor.go
```diff
@@ -0,0 +1,82 @@
...
+		AllowPartial: true,
+		EmitUnknown: true,
+	}.Format(message)
+}
+
+func tfLogRequest(ctx context.Context, req connect.AnyRequest) {
+	reqFields := tfLogBaseFields(req)
+
+	// This is tricky, because AnyRequest does not expose the underlying proto message directly, but always has it (for unary requests).
+	if reqMsg, ok := reflect.ValueOf(req).Elem().FieldByName("Msg").Interface().(proto.Message); ok {
```
idodod: any reason we can't use https://pkg.go.dev/google.golang.org/protobuf/reflect/protoreflect instead?

sosheskaz: It doesn't buy us much for this, since `req` is not a `proto.Message`. I don't think that this is the purpose of protoreflect either, which is primarily about introspecting into messages (which isn't relevant because we're deferring to `prototext` for that here). For example, side-by-side comparison of this, vs what I can do in protoreflect:

```golang
reflect.ValueOf(req).Elem().FieldByName("Msg").Interface().(proto.Message)
```

```golang
protoreflect.ValueOf(reflect.ValueOf(req).Elem().FieldByName("Msg").Interface()).Message()
```

The source code of protoreflect's `.Message()` is:

```golang
func (v Value) Message() Message {
	switch vi := v.getIface().(type) {
	case Message:
		return vi
	default:
		panic(v.panicMessage("message"))
	}
}
```

A panic is really not acceptable here IMO; I would rather this interceptor fail in a way that allows the provider to keep running.

idodod (approved):

PR #161, opened 2025-11-13
fix: s3 lifecycle noncurrent version transitions are not applied

sosheskaz:
There was previously an issue where noncurrent lifecycle transition rules were not applied. This issue two facets:
* The changes should be applied.
* When this occurs, the provider should fail due to inconsistent state, instead of failing silently.

This change addresses both issues.

* The provider now uses the result from the GET requests it already performs, while polling for the change to be accepted, when setting state, instead of its input state.
  * This strategy avoids additional overhead and does not make any more requests than previous.
  * This means that future issues of this nature will fail the acceptance tests.
* The root cause of the improper application is that one of the functions, which is used in marshalling lifecycle rules from the resource model into the PUT request in a deterministic way, was dropping lifecycle rules.
  * Because the _model_, was being saved back to state (instead of the response from `GetLifecycleConfiguration`), the _actual_ state of the lifecycle configuration was never actually saved back to state. Terraform can automatically detect these discrepancies.

### DIFF
--- a/coreweave/object_storage/resource_bucket_lifecycle_configuration.go
+++ b/coreweave/object_storage/resource_bucket_lifecycle_configuration.go
@@ -324,7 +324,7 @@ func parseISO8601(s string) time.Time {
 	return t
 }
 
-func expandRules(ctx context.Context, in []LifecycleRuleModel) []s3types.LifecycleRule {
+func expandRules(ctx context.Context, in []LifecycleRuleModel) []s3types.LifecycleRule { //nolint:gocyclo
 	out := make([]s3types.LifecycleRule, 0, len(in))
 	for _, r := range in {
 		rule := s3types.LifecycleRule{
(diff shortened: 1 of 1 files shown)


coreweave/object_storage/resource_bucket_lifecycle_configuration.go
```diff
@@ -787,9 +802,11 @@ func (r *BucketLifecycleResource) Update(ctx context.Context, req resource.Updat
 	}
 
 	// wait for lifecycle config  to be read back from s3 API since it is not guaranteed to propagate immediately
-	if err := waitForLifecycleConfig(ctx, s3c, data.Bucket.ValueString(), *lifecycleConfig); err != nil {
+	if result, err := waitForLifecycleConfig(ctx, s3c, data.Bucket.ValueString(), *lifecycleConfig); err != nil {
 		handleS3Error(err, &resp.Diagnostics, data.Bucket.ValueString())
 		return
+	} else {
+		data.Rule = flattenLifecycleRules(result.Rules)
```
sosheskaz: comment why

ADustyOldMuffin (approved):

PR #171, opened 2026-01-09
feat(ckscluster): support node_port_range update operations

AndresRCB:
`node_port_range` could only be set on creation; however, the CoreWeave API supports update operations as well:
- The NodePort Range can be updated to values that include the old NodePort Range only
- Trying to update NodePort Range to values that don't include the old range will force a replace operation
- Added test to validate this behavior
- Had to move in-line functions out of `Schema` in the cks_cluster resource because the linter failed tests due to "cyclomatic complexity"

### DIFF
--- a/coreweave/cks/resource_cluster_test.go
+++ b/coreweave/cks/resource_cluster_test.go
@@ -210,11 +210,14 @@ func generateResourceNames(clusterNamePrefix string) resourceNames {
 }
 
 type testStepConfig struct {
-	TestName         string
-	Resources        resourceNames
-	ConfigPlanChecks resource.ConfigPlanChecks
-	vpc              networking.VpcResourceModel
-	cluster          cks.ClusterResourceModel
+	TestName                string
+	Resources               resourceNames
+	ConfigPlanChecks        resource.ConfigPlanChecks
+	vpc                     networking.VpcResourceModel
+	cluster                 cks.ClusterResourceModel
+	PlanOnly                bool
+	ExpectNonEmptyPlan      bool
+	ImportStateVerifyIgnore []string
 }
 
 func stringOrNull(s types.String) knownvalue.Check {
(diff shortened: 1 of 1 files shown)


coreweave/cks/resource_cluster.go
```diff
@@ -391,6 +391,10 @@ func (c *ClusterResourceModel) ToUpdateRequest(ctx context.Context) *cksv1beta1.
 		},
 	}
 
+	if np := c.NodePorts(); np != nil {
+		req.Network.ServiceNodePortRange = np
+	}
```
sosheskaz: Could this be simplified to `req.Network.ServiceNodePortRange = c.NodePorts()`, or inlined into the initial value?

coreweave/cks/resource_cluster.go
```diff
@@ -427,6 +431,96 @@ func (r *ClusterResource) Metadata(ctx context.Context, req resource.MetadataReq
...
+	prior := []types.String{}
+	planned := []types.String{}
+
+	if diag := req.StateValue.ElementsAs(ctx, &prior, false); diag.HasError() {
+		resp.Diagnostics = diag
+		return
+	}
+
+	if diag := req.PlanValue.ElementsAs(ctx, &planned, false); diag.HasError() {
+		resp.Diagnostics = diag
```
sosheskaz: `resp.Diagnostics.Append(diag...)`

coreweave/cks/resource_cluster_test.go
```diff
@@ -493,7 +528,18 @@ func TestClusterResource(t *testing.T) {
 			cluster:   *update,
 		}),
 		createClusterTestStep(ctx, t, testStepConfig{
-			TestName: "requires replace",
+			TestName: "requires replace on node_port_range shrink",
+			ConfigPlanChecks: resource.ConfigPlanChecks{
+				PreApply: []plancheck.PlanCheck{
+					plancheck.ExpectResourceAction(config.FullResourceName, plancheck.ResourceActionDestroyBeforeCreate),
+				},
+			},
```
sosheskaz: Suggestion:

CKS Create operations can be expensive in terms of time, and this test is running a lot in sequence. It is possible to test the "RequiresReplace" functionality by using "plan only mode". There are some extra steps required to do that. Would that make sense for this case?

sosheskaz (approved):

PR #223, opened 2026-01-26
feat(networking): support new host_prefixes field, deprecate host_prefix field

sosheskaz:
This change adds a `host_prefixes` field to VPC, which offers some significant configuration options relative to the `host_prefix` field. It supports multiple prefixes, non-primary prefixes, subnet allocations, and address assignment strategies.

Backwards compatibility is maintained with the host_prefix field, but it is now considered deprecated.

There are various other changes to this module:
* Testing was expanded, and state checks made stricter and more ubiquitous in the touched code.
* Additional validation is done at plan-time, to detect doomed changes prior to apply.
* `protovalidate` was updated in service of additional plan-time validation using the networking protobufs.
* Error propagation was improved. This most likely did not fix any bugs, but it should make them less likely in future.

### DIFF
--- a/coreweave/networking/resource_vpc.go
+++ b/coreweave/networking/resource_vpc.go
@@ -519,6 +519,7 @@ func (r *VpcResource) Schema(ctx context.Context, req resource.SchemaRequest, re
 				Optional:            true,
 				Computed:            true,
 				MarkdownDescription: "An IPv4 CIDR range used to allocate host addresses when booting compute into a VPC.\nThis CIDR must be have a mask size of /18. If left unspecified, a Zone-specific default value will be applied by the server.\nThis field is immutable once set.",
+				DeprecationMessage:  "`host_prefix` is deprecated. Use `host_prefixes` instead. The field will be removed in a future version. The equivalent expression for a given resource may be found by refreshing state and running `terraform state show coreweave_networking_vpc.example`.",
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplaceIfConfigured(),
 					stringplanmodifier.UseStateForUnknown(), // required for the resource to work as expected when the value is computed instead of specified
@@ -561,7 +562,7 @@ func (r *VpcResource) Schema(ctx context.Context, req resource.SchemaRequest, re
 								"gateway_address_policy": schema.StringAttribute{
 									Optional:            true,
 									Computed:            true,
-									Default:             stringdefault.StaticString("UNSPECIFIED"),
+									Default:             stringdefault.StaticString(networkingv1beta1.IPAddressManagementPolicy_UNSPECIFIED.String()),
 									MarkdownDescription: fmt.Sprintf("Describes which IP address from the prefix is allocated to the network gateway. Must be one of: %s.", enumMarkdownValues(networkingv1beta1.IPAddressManagementPolicy_GatewayAddressPolicy_name, false)),
 								},
 							},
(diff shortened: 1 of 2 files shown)


coreweave/networking/resource_vpc.go
```diff
@@ -130,21 +278,18 @@ func (v *VpcEgressResourceModel) ToProto() *networkingv1beta1.Egress {
...
-
-func (v *VpcResourceModel) Set(vpc *networkingv1beta1.VPC) {
-	if vpc == nil {
-		return
-	}
+	Id           types.String             `tfsdk:"id"`
+	Zone         types.String             `tfsdk:"zone"`
+	Name         types.String             `tfsdk:"name"`
+	HostPrefix   types.String             `tfsdk:"host_prefix"`
+	HostPrefixes types.Set                `tfsdk:"host_prefixes"`
```
AndresRCB: Does this need to be a `types.Set` instead of `[]HostPrefixResourceModel` because at least one field in `HostPrefixResourceModel` is computed?

asimeth-cw:
I think the example in `examples/resources/coreweave_networking_vpc/resource.tf` should be updated to match the new field. (Or maybe leave that one alone and add a new example)

AndresRCB (dismissed):
Just questions and comments, so approving

coreweave/networking/resource_vpc.go
asimeth-cw: Basically, yeah. Or the entire thing could be computed (I think). To maintain backwards compatibility, it's possible to only specify the old `HostPrefix` field and not the `HostPrefixes` field. At at that point, the value of `HostPrefixes` is Unknown, and `[]HostPrefixResourceModel` can't represent an Unknown value.

sosheskaz: Yes.

For `Optional|Computed` complex types, terraform needs to be able to differentiate at plan-time between concrete states, null states, and unknown states. A pointer value or slice value cannot do this, which necessitates using the TF-native types and converting.

There can be more subtlety there, but that's the main constraint we run into here.

coreweave/networking/resource_vpc.go
```diff
@@ -285,20 +520,53 @@ func (r *VpcResource) Schema(ctx context.Context, req resource.SchemaRequest, re
...
-						if req.StateValue.ValueString() != req.PlanValue.ValueString() {
-							resp.Diagnostics.AddWarning("host_prefix is immutable, changing this value will force a replacement", fmt.Sprintf("cannot change existing host_prefix '%s' to '%s'", req.StateValue.ValueString(), req.PlanValue.ValueString()))
-						}
-
-						if resp.Diagnostics.WarningsCount() > 0 {
-							resp.RequiresReplace = true
-						}
-					}, "", ""),
+					stringplanmodifier.RequiresReplaceIfConfigured(),
+					stringplanmodifier.UseStateForUnknown(), // required for the resource to work as expected when the value is computed instead of specified
```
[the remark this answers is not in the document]
brokedba: @sosheskaz  I run few tests on v0.10.0 and I confirm that it contains breaking changes. 

Variable-based configs that worked for us in v0.9 now fail because `ValidateConfig()` doesn't handle Unknown values for `vpc_prefixes`/`name` despite your comment addressing this exact issue for `host_prefix`.

As you said above, Terraform needs to "differentiate at plan-time between concrete states, null states, and unknown states" and that `UseStateForUnknown()` is "required for the resource to work as expected when the value is computed instead of specified" . But the fix was only applied  to `host_prefix`, not to `vpc_prefixes[].value` or `name` fields.

```HCL
│ validation errors:
│  - name: value length must be at least 1 characters
│  - name: value does not match regex pattern `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
│  - vpc_prefixes[0].value: value is empty, which is not a valid IP prefix
│  - vpc_prefixes[1].value: value is empty, which is not a valid IP prefix
│  - vpc_prefixes[2].value: value is empty, which is not a valid IP prefix`
```

Result: `.ValueString()` on Unknown variables returns empty string during validation, triggering false proto validation errors. 
Please see more details and fix suggestions in issue #232 I created for this bug.


PS:  I'm not familiar with go but it took me few hours to  spot the culprit (instead of reverting back to 0.9).

PR #279, opened 2026-03-03
feat(inference): Add inference deployment, gateway, and capacity claim   resources

scotchneat:
Adds the Managed Inference resource family to the provider.

## What's new

- **Resources**
  - `coreweave_inference_deployment` — deployments with runtime/autoscaling/traffic, lifecycle polling
  - `coreweave_inference_gateway` — auth (CoreWeave / W&B) and body/header/path routing
  - `coreweave_inference_capacity_claim` — reserved-capacity claims, CUSTOMER and SERVERLESS modes
- **Data sources** — `coreweave_inference_parameters`, `coreweave_inference_gateway_parameters`, `coreweave_inference_capacity_claim_parameters` surface valid zones, instance types, runtime versions, and gateway IDs
- **Acceptance tests** per resource (lifecycle + import) plus `TestAccInferenceReservedCapacity` covering the full `capacity_claim → gateway → deployment` chain with `capacity_classes = ["RESERVED"]`
- Sweepers for all three resource types, with ordering so deployments sweep before gateways/claims

## Test plan

- [x] `make testacc` passes against the STAGING env with `SUITES=inference`
- [x] Import verify passes for all three resources
- [x] `make generate` produces no drift
- [ ]  Verify docs reference urls (in schema `MarkdownDescription` properties) with docs team


coreweave/inference/resource_inference_gateway.go
```diff
@@ -0,0 +1,862 @@
...
+		ec := &inferencev1.EndpointConfiguration{}
+		if !m.EndpointConfiguration.AdditionalDNS.IsNull() && !m.EndpointConfiguration.AdditionalDNS.IsUnknown() {
+			dns := []string{}
+			diagnostics.Append(m.EndpointConfiguration.AdditionalDNS.ElementsAs(ctx, &dns, false)...)
+			if diagnostics.HasError() {
+				return gatewayFields{}, diagnostics
+			}
+			ec.AdditionalDns = dns
+		}
+		f.EndpointConfiguration = ec
```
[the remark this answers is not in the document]
scotchneat: @kulkarnisamr Can you verify what the API behavior is if it receives an empty/non-nil endpoint configuration?
Does it honor the request and clear the existing endpoint settings?

coreweave/inference/resource_inference_gateway.go
```diff
@@ -0,0 +1,862 @@
...
+	full := "API_TYPE_" + s
+	if val, ok := inferencev1.BodyBasedRouting_APIType_value[full]; ok {
+		return inferencev1.BodyBasedRouting_APIType(val)
+	}
+	return inferencev1.BodyBasedRouting_API_TYPE_UNSPECIFIED
+}
+
+// apiTypeToString converts a proto API type enum value to the user-facing string (e.g. "OPENAI").
+func apiTypeToString(at inferencev1.BodyBasedRouting_APIType) string {
+	s := at.String()
```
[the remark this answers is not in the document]
scotchneat: This helper is a divergent pattern from the rest of the provider, so I'd rather make the inference resource follow the patterns already set.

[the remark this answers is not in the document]
scotchneat: Refactored inference to match existing networking pattern.

coreweave/inference/resource_inference_gateway.go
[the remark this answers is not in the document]
kulkarnisamr: if it receives an empty endpoint configuration it will still get an automatically generated (basedomain = `cwinference.com`) dns record, so that will be returned. example:
```
> ./infr-cli gateway get "${GW_ID}" --insecure
Handling connection for 18443
FIELD            VALUE
ID               25d4a807-31e7-45eb-92db-a6e5d2281fca
Name             cw7634-dev-gw
Organization     cw7364
Zones            RNO2A
Authentication   CoreWeave
Routing          Body(API_TYPE_OPENAI)
Status           STATUS_UNSPECIFIED
Endpoints        https://cw7634-dev-gw.cw7364.gw.cwinference.com
Created          2026-04-23 20:29:05
Updated          2026-04-23 20:29:05
```

for non-nil endpoint configuration i.e. custom DNS endpoints provided by the customer, it will get returned along with this automatically generated one (merged list of all endpoints).

ref: https://github.com/coreweave/infr-api/pull/427

docs/resources/inference_gateway.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Example Usage
+
+```terraform
+resource "coreweave_inference_gateway" "example" {
+  name  = "my-gateway"
+  zones = ["US-EAST-04A"]
+
+  auth = {
+    core_weave = {}
```
sosheskaz: I think `coreweave` is the more consistent styling, and we should stick with that unless there's a broader style directive.

coreweave/inference/resource_inference_gateway.go
```diff
@@ -0,0 +1,842 @@
...
+			},
+			"endpoints": schema.ListAttribute{
+				Computed:            true,
+				ElementType:         types.StringType,
+				MarkdownDescription: "The endpoint URIs for the gateway.",
+			},
+			"name": schema.StringAttribute{
+				Required:            true,
+				MarkdownDescription: "The human-readable name of the gateway.",
+			},
```
sosheskaz, line 166: The `name` attribute lacks a `RequiresReplace` plan modifier — the deployment ([line 163](https://github.com/coreweave/terraform-provider-coreweave/blob/ba4b4c10c723096a9aa3a0ca99813550fb1ec68a/coreweave/inference/resource_inference_deployment.go#L163)) and capacity_claim ([line 126](https://github.com/coreweave/terraform-provider-coreweave/blob/ba4b4c10c723096a9aa3a0ca99813550fb1ec68a/coreweave/inference/resource_inference_capacity_claim.go#L126)) both apply `stringplanmodifier.RequiresReplace()` on their `name`. If the gateway API rejects rename via `UpdateGateway`, a renamed config will fail at apply time rather than triggering destroy/recreate.

[Quickstart](https://coreweave.atlassian.net/wiki/spaces/CLOUDPLAT/pages/497779629/Terraform+Provider+Developer+Quickstart): *"When an argument is viable for Create, but not for Update operations, the RequiresReplace modifier must be used. This also applies to any immutable fields."*

Worth verifying the API contract; if name is immutable, add the modifier here.

examples/resources/coreweave_inference_deployment/resource.tf
```diff
@@ -0,0 +1,38 @@
+# Look up available parameters first (optional but recommended).
```
sosheskaz, line 1: This resource implements `ImportState` ([resource_inference_deployment.go:551-553](https://github.com/coreweave/terraform-provider-coreweave/blob/ba4b4c10c723096a9aa3a0ca99813550fb1ec68a/coreweave/inference/resource_inference_deployment.go#L551-L553)) but `examples/resources/coreweave_inference_deployment/import.sh` is missing. The `coreweave_inference_capacity_claim` and `coreweave_inference_gateway` examples both ship one. Without it, `terraform-plugin-docs` will not emit an Import section in the generated [`docs/resources/inference_deployment.md`](https://github.com/coreweave/terraform-provider-coreweave/blob/ba4b4c10c723096a9aa3a0ca99813550fb1ec68a/docs/resources/inference_deployment.md).

[Quickstart](https://coreweave.atlassian.net/wiki/spaces/CLOUDPLAT/pages/497779629/Terraform+Provider+Developer+Quickstart): *"All importable resources should have an example import at examples/resources/<resource name>/import.sh."*

coreweave/inference/resource_inference_deployment.go
```diff
@@ -0,0 +1,837 @@
...
+					},
+					"gpu_count": schema.Int64Attribute{
+						Required:            true,
+						MarkdownDescription: "Number of GPUs per instance. Must be one of: 1, 2, 4, 8, 16.",
+						Validators: []validator.Int64{
+							int64validator.OneOf(1, 2, 4, 8, 16),
+						},
+					},
+				},
+			},
```
sosheskaz, line 243: `instance_type` and `gpu_count` likely cannot be hot-swapped on a running deployment. If the API treats them as immutable post-create, both should carry `stringplanmodifier.RequiresReplace()` / `int64planmodifier.RequiresReplace()` (or this should be confirmed mutable in the API contract).

The acceptance test currently exercises only `disabled`, autoscaling, and traffic updates ([resource_inference_deployment_test.go:426+](https://github.com/coreweave/terraform-provider-coreweave/blob/ba4b4c10c723096a9aa3a0ca99813550fb1ec68a/coreweave/inference/resource_inference_deployment_test.go#L426)), so any update-immutability rejection on these fields would not be caught by CI.

coreweave/inference/resource_inference_capacity_claim.go
```diff
@@ -0,0 +1,536 @@
...
+					"zones": schema.SetAttribute{
+						ElementType:         types.StringType,
+						Required:            true,
+						MarkdownDescription: "The availability zones where the capacity claim may use resources from (e.g. `US-WEST-04A`). Case insensitive. At least one is required.",
+						Validators: []validator.Set{
+							setvalidator.SizeAtLeast(1),
+						},
+					},
+				},
+			},
```
sosheskaz, line 181: Capacity reservations are typically immutable after creation. If `instance_id`, `capacity_type`, and `zones` cannot be updated server-side, all three should carry `RequiresReplace`. Same concern as the deployment `resources` block — the update test path doesn't currently exercise these fields, so an API rejection would surface only when users actually attempt the change.

Worth pinning down which of the four fields here are mutable per the inference API spec, and adding `RequiresReplace` to the rest.

coreweave/inference/resource_inference_deployment.go
```diff
@@ -0,0 +1,837 @@
...
+			"conditions": schema.ListNestedAttribute{
+				Computed:            true,
+				MarkdownDescription: "Detailed status conditions for the deployment.",
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"type":             schema.StringAttribute{Computed: true},
+						"status":           schema.StringAttribute{Computed: true},
+						"last_update_time": schema.StringAttribute{Computed: true},
+						"reason":           schema.StringAttribute{Computed: true},
+						"message":          schema.StringAttribute{Computed: true},
```
sosheskaz: The five `conditions` sub-attributes (`type`, `status`, `last_update_time`, `reason`, `message`) have no `MarkdownDescription`. The same gap exists in the gateway and capacity_claim resources, plus the parameters datasources (`runtime_versions.versions`, `runtime_config_options.allowed_keys`, `zone_instance_types.instance_ids`) — about nine sites total.

[Quickstart](https://coreweave.atlassian.net/wiki/spaces/CLOUDPLAT/pages/497779629/Terraform+Provider+Developer+Quickstart): *"All attributes should have a description associated with them, explaining the effect that the attribute has."* These descriptions feed directly into the generated docs.

coreweave/inference/resource_inference_capacity_claim.go
```diff
@@ -0,0 +1,536 @@
...
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	createResp, err := r.client.CreateCapacityClaim(ctx, connect.NewRequest(createReq))
+	if err != nil {
+		coreweave.HandleAPIError(ctx, err, &resp.Diagnostics)
+		return
+	}
```
sosheskaz, line 220: Low: This is probably a good change but we may not want to block on it.

Neither `InferenceDeploymentResource` nor `InferenceCapacityClaimResource` runs `protovalidate.Validate(createReq)` (the gateway only adds `ConfigValidators` for `ExactlyOneOf`). The [Quickstart](https://coreweave.atlassian.net/wiki/spaces/CLOUDPLAT/pages/497779629/Terraform+Provider+Developer+Quickstart) recommends it as *"a good way to get inexpensive, automatic client-side validation"* via `ValidateConfig`.

Verified that this is a systemic repo gap — no resource in `coreweave/` currently calls `protovalidate.Validate` (only one comment-only reference in `networking/resource_vpc_test.go:687`). Treat as a low-priority follow-up rather than a blocker; flagging here so we don't ship three more resources without it.

Side note: the deployment lacks `ConfigValidators` entirely, so cross-field constraints like `autoscaling.min <= autoscaling.max` are validated only server-side.

coreweave/inference/resource_inference_deployment.go
```diff
@@ -0,0 +1,837 @@
...
+
+	updateResp, err := r.client.UpdateDeployment(ctx, connect.NewRequest(updateReq))
+	if err != nil {
+		coreweave.HandleAPIError(ctx, err, &resp.Diagnostics)
+		return
+	}
+
+	deploymentID := updateResp.Msg.Deployment.GetSpec().GetId()
+
+	conf := retry.StateChangeConf{
```
sosheskaz, line 488: After `UpdateDeployment` returns successfully, this jumps straight into `WaitForStateContext` without persisting the partial response into state. If polling times out (the `Timeout` here is 20 minutes) or the context is cancelled, state still reflects the *pre-update* values — so a subsequent plan won't realize the update is in flight on the server.

Compare with `Create` ([lines 344-350](https://github.com/coreweave/terraform-provider-coreweave/blob/ba4b4c10c723096a9aa3a0ca99813550fb1ec68a/coreweave/inference/resource_inference_deployment.go#L344-L350)), which deliberately writes intermediate state before polling. Risk is lower for Update than Create (the resource already exists in state) but a pre-poll save here would be a small consistency win.

Same pattern in gateway and capacity_claim Update functions.

coreweave/inference/resource_inference_deployment_test.go
```diff
@@ -0,0 +1,493 @@
...
+    precondition {
+      condition     = local.preferred_instance == "" || contains(local.available_instances, local.preferred_instance)
+      error_message = "INFR_INSTANCE_ID=\"${local.preferred_instance}\" is not present in inference parameters; available: ${jsonencode(local.available_instances)}"
+    }
+  }
+}
+`, preferredZone, preferredInstance, name, name)
+}
+
+func TestAccInferenceDeployment(t *testing.T) {
```
sosheskaz: The [Quickstart](https://coreweave.atlassian.net/wiki/spaces/CLOUDPLAT/pages/497779629/Terraform+Provider+Developer+Quickstart#Acceptance-Tests) prescribes:

```go
func TestInferenceDeployment(t *testing.T) {
  t.Parallel()
  t.Run("lifecycle", func(t *testing.T) { resource.ParallelTest(t, ...) })
}
```

so a single resource's full suite (unit + acceptance) is reachable via `TF_ACC=1 go test -run TestInferenceDeployment ./coreweave/inference`. With the current naming, `-run TestInferenceDeployment` catches only the unit tests (`TestInferenceDeploymentResource_Schema`, etc.) and skips the lifecycle test — `TestAccInferenceDeployment` lives under a separate prefix.

CI uses `make testacc SUITES=inference` so practical impact is low, but local TDD ergonomics suffer. Same pattern in the gateway and capacity_claim test files. Separately: the use of `resource.Test` (vs. `ParallelTest`) with `//nolint:forbidigo` is a legitimate deviation given staging zone constraints — no concern there.

coreweave/schema_docs.go
```diff
@@ -0,0 +1,28 @@
...
+	"fmt"
+	"slices"
+	"strings"
+)
+
+// EnumMarkdownValues renders the values of an int-keyed name map as a
+// backtick-quoted, comma-separated list sorted by key. When dropZero is true
+// the zero-key entry is omitted.
+func EnumMarkdownValues(m map[int32]string, dropZero bool) string {
+	// the values should be sorted by their enum (string) value, not the string value.
```
sosheskaz: Inherited a stray `(string)` during the move from `resource_vpc.go`. `keys` is `[]int32` and `slices.Sort(keys)` sorts numerically — the pre-extraction comment (`// sorted by their enum value, not the string value.`) was correct.

coreweave/inference/resource_inference_capacity_claim_test.go
```diff
@@ -0,0 +1,324 @@
...
+)
+
+const testInstanceID = "gb200-4x"
+
+// --- Unit tests ---
+
+func TestInferenceCapacityClaimResource_Schema(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
```
sosheskaz: CLAUDE.md says: *"When writing tests, one should typically use `t.Context()`, not `context.Background()`."* This file uses `t.Context()` correctly at line 141 (`TestToCreateCapacityClaimRequest_Fields`), so the convention is mixed within the PR. Other instance in this file: line 120 (`m.Resources.Zones.ElementsAs(context.Background(), ...)`).

coreweave/inference/resource_inference_deployment_test.go
```diff
@@ -0,0 +1,493 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
+)
+
+// --- Unit tests for pure helper functions ---
+
+func TestInferenceDeploymentResource_Schema(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
```
sosheskaz: CLAUDE.md says: *"When writing tests, one should typically use `t.Context()`, not `context.Background()`."* Other instances in this file: line 123 (`TestToCreateRequest_OptionalFields`) and line 184 (`TestToUpdateRequest_Fields`).

coreweave/inference/resource_inference_gateway_test.go
```diff
@@ -0,0 +1,632 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
+)
+
+// --- Unit tests ---
+
+func TestInferenceGatewayResource_Schema(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
```
sosheskaz: CLAUDE.md says: *"When writing tests, one should typically use `t.Context()`, not `context.Background()`."* The PR uses `t.Context()` correctly elsewhere (e.g., `TestToCreateGatewayRequest_*`), so the convention is mixed within the same file.

coreweave/inference/resource_inference_deployment.go
```diff
@@ -0,0 +1,837 @@
...
+			inferencev1.Status_STATUS_CREATING.String(),
+			inferencev1.Status_STATUS_UNSPECIFIED.String(),
+		},
+		Target: []string{inferencev1.Status_STATUS_READY.String()},
+		Refresh: func() (interface{}, string, error) {
+			getResp, err := r.client.GetDeployment(ctx, connect.NewRequest(&inferencev1.GetDeploymentRequest{
+				Id: deploymentID,
+			}))
+			if err != nil {
+				tflog.Error(ctx, "failed to poll deployment", map[string]interface{}{"error": err})
```
sosheskaz: `tflog.Error` JSON-marshals each map value. The `error` interface has no exported fields, so passing `err` here yields `{"error": {}}` in the structured log — the actual message is dropped. The Update path at line 458 correctly uses `err.Error()`. Same divergence in `gateway.go` (315 vs 407) and `capacity_claim.go` (225 vs 319).

coreweave/inference/resource_inference_gateway.go
```diff
@@ -0,0 +1,842 @@
...
+			inferencev1.Status_STATUS_CREATING.String(),
+			inferencev1.Status_STATUS_UNSPECIFIED.String(),
+		},
+		Target: []string{inferencev1.Status_STATUS_READY.String()},
+		Refresh: func() (interface{}, string, error) {
+			getResp, err := r.client.GetGateway(ctx, connect.NewRequest(&inferencev1.GetGatewayRequest{
+				Id: gatewayID,
+			}))
+			if err != nil {
+				tflog.Error(ctx, "failed to poll gateway", map[string]interface{}{"error": err})
```
sosheskaz: `tflog.Error` JSON-marshals each map value. Passing `err` (interface) yields `{"error": {}}` — the message is dropped. Update at line 407 uses `err.Error()` correctly.

coreweave/inference/resource_inference_capacity_claim.go
```diff
@@ -0,0 +1,536 @@
...
+		},
+		Target: []string{
+			inferencev1.Status_STATUS_READY.String(),
+		},
+		Refresh: func() (interface{}, string, error) {
+			getResp, err := r.client.GetCapacityClaim(ctx, connect.NewRequest(&inferencev1.GetCapacityClaimRequest{
+				Id: claimID,
+			}))
+			if err != nil {
+				tflog.Error(ctx, "failed to poll capacity claim", map[string]interface{}{"error": err})
```
sosheskaz: `tflog.Error` JSON-marshals each map value. Passing `err` (interface) yields `{"error": {}}` — the message is dropped. Update at line 319 uses `err.Error()` correctly.

coreweave/inference/resource_inference_deployment.go
```diff
@@ -0,0 +1,837 @@
...
+				},
+			},
+			"gateway_ids": schema.SetAttribute{
+				ElementType:         types.StringType,
+				Required:            true,
+				MarkdownDescription: "The gateway IDs to associate the deployment with. At least one is required. Order is not significant.",
+				Validators: []validator.Set{
+					setvalidator.SizeAtLeast(1),
+					setvalidator.ValueStringsAre(
+						stringvalidator.RegexMatches(uuidPattern, "must be a valid UUID"),
```
sosheskaz: The `gateway_ids` validator uses a strict UUID regex on user-supplied IDs. If the API ever returns IDs in a different format (or future versions add prefixes), every existing config breaks at plan time. Consider relaxing this to a non-empty string check, since the API is the source of truth for the format.

coreweave/inference/resource_inference_capacity_claim_test.go
```diff
@@ -0,0 +1,324 @@
...
+func TestAccInferenceCapacityClaim(t *testing.T) {
+	name := fmt.Sprintf("%scc-%x", AcceptanceTestPrefix, rand.IntN(100000))
+	fullResourceName := "coreweave_inference_capacity_claim.test"
+	preferredZone := preferredInferenceZone()
+	preferredInstance := preferredInferenceInstanceType()
+
+	// Inference acceptance tests run sequentially (resource.Test, not
+	// resource.ParallelTest) because the staging environment has limited
+	// per-zone capacity; parallelism causes allocation failures.
+	//nolint:forbidigo // see comment above
```
sosheskaz: Same forbidigo suppression pattern as in `resource_inference_gateway_test.go`. This is a single-resource lifecycle test — there is no apparent reason it cannot run as `resource.ParallelTest`. CLAUDE.md treats lint suppressions as destructive actions; please justify with a comment or switch to `resource.ParallelTest`.

coreweave/inference/resource_inference_deployment.go
```diff
@@ -0,0 +1,837 @@
...
+					"concurrency": schema.Int64Attribute{
+						Optional:            true,
+						MarkdownDescription: "Concurrency per instance target (≥1). Controls latency vs throughput tradeoffs.",
+						Validators: []validator.Int64{
+							int64validator.AtLeast(1),
+						},
+					},
+				},
+			},
+			"traffic": schema.SingleNestedAttribute{
```
sosheskaz, line 310: The `traffic` block is `Required` but its only attribute (`weight`) is `Optional`/`Computed`. That forces every config to write `traffic = {}` boilerplate. Consider either making the block itself `Optional` + `Computed`, or hoisting `weight` to a top-level attribute, depending on the intended UX.

coreweave/inference/resource_inference_capacity_claim.go
```diff
@@ -0,0 +1,536 @@
...
+					stringvalidator.RegexMatches(hostnamePattern, "must be a valid hostname label"),
+				},
+			},
+			"resources": schema.SingleNestedAttribute{
+				Required:            true,
+				MarkdownDescription: "Resource configuration for the capacity claim.",
+				Attributes: map[string]schema.Attribute{
+					"instance_id": schema.StringAttribute{
+						Required:            true,
+						MarkdownDescription: "The instance type to reserve by ID specifier (e.g. `gb200-4x`). Case insensitive.",
```
sosheskaz: Doc says `Case insensitive.` but the schema does not normalize the value (no plan modifier, no validator). Either add a plan modifier that lowercases on read/write to honor the doc, or drop the `Case insensitive` claim. Otherwise users will see spurious diffs when the API returns a different case than they wrote.

coreweave/inference/resource_inference_capacity_claim.go
```diff
@@ -0,0 +1,536 @@
...
+						},
+					},
+					"capacity_type": schema.StringAttribute{
+						Required:            true,
+						MarkdownDescription: fmt.Sprintf("The capacity type for the capacity claim. Must be one of: %s.", coreweave.EnumMarkdownValues(inferencev1.CapacityType_name, true)),
+					},
+					"zones": schema.SetAttribute{
+						ElementType:         types.StringType,
+						Required:            true,
+						MarkdownDescription: "The availability zones where the capacity claim may use resources from (e.g. `US-WEST-04A`). Case insensitive. At least one is required.",
```
sosheskaz: Same `Case insensitive` claim without normalization as on `instance_id`. Either normalize via a plan modifier or remove the claim from the description to keep the doc and behavior aligned.

coreweave/inference/resource_inference_gateway.go
```diff
@@ -0,0 +1,842 @@
...
+		return
+	}
+
+	updateReq, diags := toUpdateGatewayRequest(ctx, &data)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	updateResp, err := r.client.UpdateGateway(ctx, connect.NewRequest(updateReq))
```
sosheskaz, line 411: Same pattern as the Create path: after `UpdateGateway` returns successfully, this calls `WaitForGatewayActive` without first persisting the intermediate state. If polling fails or times out, the user is left with state that does not reflect the in-flight update. Consider an intermediate `resp.State.Set` between the Update RPC and the polling loop, mirroring the Create flow described in CLAUDE.md (`Why save state twice?`).

coreweave/inference/resource_inference_capacity_claim.go
```diff
@@ -0,0 +1,536 @@
...
+		return
+	}
+
+	updateReq, diags := toUpdateCapacityClaimRequest(ctx, &data)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	updateResp, err := r.client.UpdateCapacityClaim(ctx, connect.NewRequest(updateReq))
```
sosheskaz, line 320: Update path lacks an intermediate `resp.State.Set` between the API call and the polling loop. CLAUDE.md (`Why save state twice?`) calls this out for Create — the same reasoning applies to Update. If the poll times out, the user is stuck with stale state for an in-flight change.

coreweave/inference/resource_inference_gateway.go
```diff
@@ -0,0 +1,842 @@
...
+						Attributes:          map[string]schema.Attribute{},
+					},
+					"weights_and_biases": schema.SingleNestedAttribute{
+						Optional:            true,
+						MarkdownDescription: "Use Weights & Biases authentication.",
+						Attributes: map[string]schema.Attribute{
+							"api_key": schema.StringAttribute{
+								Optional:            true,
+								Sensitive:           true,
+								MarkdownDescription: "The organization API key for Weights & Biases. Required if `server_url` is set.",
```
sosheskaz, line 191: Doc string says "Required if `server_url` is set" but no validator enforces this — an invalid config will only fail at apply time. Consider adding a `resourcevalidator.RequiredTogether` (or equivalent) so the validation runs at plan time. CLAUDE.md describes this in the `Multi-Layer Validation Strategy` section: schema-level, config-level, and proto-level checks.

coreweave/inference/resource_inference_deployment.go
```diff
@@ -0,0 +1,837 @@
...
+	}, diags
+}
+
+// setFromDeployment populates all fields on the model from a proto Deployment response.
+// For Optional (non-Computed) fields, null is preserved when the plan/state was null and the
+// API returns the default (zero/empty) value — matching the CKS pattern for optional fields.
+// ensureDeploymentNested initializes each required nested attribute on the model
+// if it is nil, so the model arriving from an ID-only import state is safe to
+// write into. Each initializer seeds optional inner fields as null so null
+// preservation checks in setFromDeployment behave correctly on first import.
```
sosheskaz: These comment lines describe `setFromDeployment` (defined at line 709) but they sit immediately above `ensureDeploymentNested` with no blank line in between, making them read as documentation for the wrong function. Add a blank line between the doc comment and `ensureDeploymentNested`, or move the doc block down to where `setFromDeployment` is defined.

captainshar:
Hi @scotchneat — picking up the docs verification on the open test-plan item. Once #2856 merges, these are the canonical doc URLs you can reference from `MarkdownDescription` schema properties:

**Conceptual / how-to**

- https://docs.coreweave.com/products/inference — overview
- https://docs.coreweave.com/products/inference/getting-started — end-to-end walkthrough (gateway → deployment → request)
- https://docs.coreweave.com/products/inference/concepts/gateways — auth, routing strategies, traffic splitting
- https://docs.coreweave.com/products/inference/concepts/gateways#weights-and-biases-authentication — W&B SaaS auth (preview)
- https://docs.coreweave.com/products/inference/concepts/models — runtime, GPU selection, model storage
- https://docs.coreweave.com/products/inference/concepts/models#grant-inference-access-to-your-bucket — required bucket policy for the inference service account
- https://docs.coreweave.com/products/inference/concepts/scaling — autoscaling parameters
- https://docs.coreweave.com/products/inference/concepts/scaling#capacity-claims — capacity claims (RESERVED vs ON_DEMAND, CUSTOMER vs SERVERLESS)
- https://docs.coreweave.com/products/inference/dedicated — Dedicated Inference

**API reference**

- https://docs.coreweave.com/products/inference/reference/api-overview — REST/gRPC/Connect overview, status values, error format

coreweave/inference/resource_inference_deployment.go
```diff
@@ -0,0 +1,837 @@
...
+	}
+
+	if d.GetStatus().GetStatus() == inferencev1.Status_STATUS_ERROR ||
+		d.GetStatus().GetStatus() == inferencev1.Status_STATUS_FAILED {
+		resp.Diagnostics.AddError("Deployment creation failed",
+			fmt.Sprintf("Deployment entered status %s. You must destroy and recreate this resource.", d.GetStatus().GetStatus().String()))
+	}
+
+	resp.Diagnostics.Append(setFromDeployment(&data, d)...)
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
```
sosheskaz, line 430: After the `AddError` block (lines 392–396) detects `STATUS_FAILED`, execution continues into `setFromDeployment` and `resp.State.Set` — writing state while `resp.Diagnostics` already contains an error.

The `errDeploymentFailed` sentinel suggests this is intentional: persisting the FAILED state lets users inspect `conditions` without the resource disappearing from state. But the intent is undocumented and the pattern conflicts with the framework convention that `State.Set` should not be called when diagnostics contain an error. On the next `plan`, Terraform treats the persisted FAILED state as ground truth and generates a diff against it.

**Options:**
1. Add `if resp.Diagnostics.HasError() { return }` before `setFromDeployment` (standard pattern — the pre-polling `State.Set` on line 348 ensures the resource is not orphaned). Users see a clear error and re-run.
2. Keep the state write but add an explicit comment explaining the intent and the accepted tradeoff.

Same pattern in the Update path (lines 484–491), and in `resource_inference_gateway.go` (Create: 341–348, Update: 433–440) and `resource_inference_capacity_claim.go` (Create: 251–258, Update: 345–352).

coreweave/inference/resource_inference_deployment.go
```diff
@@ -0,0 +1,837 @@
...
+						MarkdownDescription: "Priority for cross-deployment scaling (0–1000). Higher values win when there is contention.",
+						Validators: []validator.Int64{
+							int64validator.Between(0, 1000),
+						},
+					},
+					"capacity_classes": schema.ListAttribute{
+						ElementType:         types.StringType,
+						Optional:            true,
+						MarkdownDescription: fmt.Sprintf("Capacity classes to use. Allowed values: %s.", coreweave.EnumMarkdownValues(inferencev1.DeploymentAutoscaling_CapacityClass_name, true)),
+					},
```
sosheskaz, line 300: `capacity_classes` has no `Validators` — plan-time enum validation was removed when the description was updated to use `EnumMarkdownValues`. The same applies to `capacity_type` in `resource_inference_capacity_claim.go` (line 146) and `api_type` in `resource_inference_gateway.go` (line 201).

Previously each field carried a proto-derived `stringvalidator.OneOf(...)` guard. That check now only exists inside `toCreateRequest`/`toUpdateRequest` at apply time (e.g. line 631). Any invalid enum value passes plan validation silently and is only rejected by the API after resource provisioning may have begun.

Consider re-adding `stringvalidator.OneOf` values derived from the proto map (with the `UNSPECIFIED=0` key excluded). For `capacity_classes`, `listvalidator.ValueStringsAre(stringvalidator.OneOf(...))` is the correct form. This restores plan-time validation independently of whether `ValidateConfig`+`protovalidate` are added.

coreweave/inference/resource_inference_deployment.go
```diff
@@ -0,0 +1,837 @@
...
+	}
+
+	if d.GetStatus().GetStatus() == inferencev1.Status_STATUS_ERROR ||
+		d.GetStatus().GetStatus() == inferencev1.Status_STATUS_FAILED {
+		resp.Diagnostics.AddError("Deployment update failed",
+			fmt.Sprintf("Deployment entered status %s. Check the `conditions` attribute for details.", d.GetStatus().GetStatus().String()))
+	}
+
+	resp.Diagnostics.Append(setFromDeployment(&data, d)...)
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
```
sosheskaz, line 534: Same pattern as Create (line 399): after `AddError` detects `STATUS_FAILED`, execution continues into `setFromDeployment` and `resp.State.Set` — writing state while diagnostics already contain an error. Add `if resp.Diagnostics.HasError() { return }` before `setFromDeployment`, or document the intent explicitly (the pre-polling `State.Set` at line 448 prevents orphaning either way).

coreweave/inference/resource_inference_capacity_claim.go
```diff
@@ -0,0 +1,536 @@
...
+		return
+	}
+
+	if cc.GetStatus().GetStatus() == inferencev1.Status_STATUS_ERROR ||
+		cc.GetStatus().GetStatus() == inferencev1.Status_STATUS_FAILED {
+		resp.Diagnostics.AddError("Capacity claim creation failed",
+			fmt.Sprintf("Capacity claim entered status %s. You must destroy and recreate this resource.", cc.GetStatus().GetStatus().String()))
+	}
+
+	resp.Diagnostics.Append(setFromCapacityClaim(&data, cc)...)
```
sosheskaz, line 280: After `AddError` detects `STATUS_FAILED` (lines 251–255), execution continues into `setFromCapacityClaim` and `resp.State.Set` — writing state while diagnostics already contain an error. Add `if resp.Diagnostics.HasError() { return }` before `setFromCapacityClaim`, or document that persisting the FAILED state for `conditions` inspection is intentional. The pre-polling `State.Set` at line 205 prevents orphaning.

coreweave/inference/resource_inference_capacity_claim.go
```diff
@@ -0,0 +1,536 @@
...
+		return
+	}
+
+	if cc.GetStatus().GetStatus() == inferencev1.Status_STATUS_ERROR ||
+		cc.GetStatus().GetStatus() == inferencev1.Status_STATUS_FAILED {
+		resp.Diagnostics.AddError("Capacity claim update failed",
+			fmt.Sprintf("Capacity claim entered status %s. Check the `conditions` attribute for details.", cc.GetStatus().GetStatus().String()))
+	}
+
+	resp.Diagnostics.Append(setFromCapacityClaim(&data, cc)...)
```
sosheskaz, line 386: Same as Create (line 257): `AddError` is called for `STATUS_FAILED` (lines 345–349) but `setFromCapacityClaim` and `resp.State.Set` run unconditionally. Add `if resp.Diagnostics.HasError() { return }` before `setFromCapacityClaim`, or document the intent.

coreweave/inference/resource_inference_gateway.go
```diff
@@ -0,0 +1,842 @@
...
+		return
+	}
+
+	if gw.GetStatus().GetStatus() == inferencev1.Status_STATUS_ERROR ||
+		gw.GetStatus().GetStatus() == inferencev1.Status_STATUS_FAILED {
+		resp.Diagnostics.AddError("Gateway creation failed",
+			fmt.Sprintf("Gateway entered status %s. You must destroy and recreate this resource.", gw.GetStatus().GetStatus().String()))
+	}
+
+	resp.Diagnostics.Append(setFromGateway(&data, gw)...)
```
sosheskaz, line 371: After `AddError` detects `STATUS_FAILED` (lines 341–345), execution continues into `setFromGateway` and `resp.State.Set` — writing state while diagnostics already contain an error. Add `if resp.Diagnostics.HasError() { return }` before `setFromGateway`, or document that persisting the FAILED state for `conditions` inspection is intentional. The pre-polling `State.Set` at line 297 prevents orphaning.

coreweave/inference/resource_inference_gateway.go
```diff
@@ -0,0 +1,842 @@
...
+		return
+	}
+
+	if gw.GetStatus().GetStatus() == inferencev1.Status_STATUS_ERROR ||
+		gw.GetStatus().GetStatus() == inferencev1.Status_STATUS_FAILED {
+		resp.Diagnostics.AddError("Gateway update failed",
+			fmt.Sprintf("Gateway entered status %s. Check the `conditions` attribute for details.", gw.GetStatus().GetStatus().String()))
+	}
+
+	resp.Diagnostics.Append(setFromGateway(&data, gw)...)
```
sosheskaz, line 475: Same as Create (line 347): `AddError` is called for `STATUS_FAILED` (lines 433–437) but `setFromGateway` and `resp.State.Set` run unconditionally. Add `if resp.Diagnostics.HasError() { return }` before `setFromGateway`, or document the intent.

coreweave/inference/resource_inference_capacity_claim.go
```diff
@@ -0,0 +1,536 @@
...
+					"instance_count": schema.Int64Attribute{
+						Required:            true,
+						MarkdownDescription: "The number of instances to reserve. Must be at least 1.",
+						Validators: []validator.Int64{
+							int64validator.AtLeast(1),
+						},
+					},
+					"capacity_type": schema.StringAttribute{
+						Required:            true,
+						MarkdownDescription: fmt.Sprintf("The capacity type for the capacity claim. Must be one of: %s.", coreweave.EnumMarkdownValues(inferencev1.CapacityType_name, true)),
```
sosheskaz: `capacity_type` has no `Validators` — plan-time enum validation was removed when the description was updated to use `EnumMarkdownValues`. Any invalid value passes plan validation silently and is only rejected by the API at apply time (see the validation logic at line 437 in `toCreateCapacityClaimRequest`). Re-add `stringvalidator.OneOf` with values derived from `inferencev1.CapacityType_name` (skipping key 0).

coreweave/inference/resource_inference_gateway.go
```diff
@@ -0,0 +1,842 @@
...
+				Required:            true,
+				MarkdownDescription: "The routing configuration for the gateway. Exactly one of `body_based`, `header_based`, or `path_based` must be specified.",
+				Attributes: map[string]schema.Attribute{
+					"body_based": schema.SingleNestedAttribute{
+						Optional:            true,
+						MarkdownDescription: "Body-based routing configuration. Routes requests based on the request body content.",
+						Attributes: map[string]schema.Attribute{
+							"api_type": schema.StringAttribute{
+								Required:            true,
+								MarkdownDescription: fmt.Sprintf("The well-known API type for routing. Must be one of: %s.", coreweave.EnumMarkdownValues(inferencev1.BodyBasedRouting_APIType_name, true)),
```
sosheskaz, line 219: `api_type` has no `Validators` — plan-time enum validation was removed when the description was updated to use `EnumMarkdownValues`. Any invalid value passes plan validation silently and is only rejected by the API at apply time (see the validation logic at line 838 in `buildGatewayFields`). Re-add `stringvalidator.OneOf` with values derived from `inferencev1.BodyBasedRouting_APIType_name` (skipping key 0).

sosheskaz (commented):
Collection type review (List vs Set) — these were missed in my initial pass due to a triage gap. Six findings about phantom-diff risk from `ListAttribute` where `SetAttribute` is appropriate, plus two pre-allocation nits.

coreweave/inference/resource_inference_gateway.go
```diff
@@ -0,0 +1,842 @@
...
+					Attributes: map[string]schema.Attribute{
+						"type":             schema.StringAttribute{Computed: true},
+						"status":           schema.StringAttribute{Computed: true},
+						"last_update_time": schema.StringAttribute{Computed: true},
+						"reason":           schema.StringAttribute{Computed: true},
+						"message":          schema.StringAttribute{Computed: true},
+					},
+				},
+			},
+			"endpoints": schema.ListAttribute{
```
sosheskaz: `endpoints` is Computed and populated entirely by the API. `ListAttribute` implies stable insertion order, but gateway endpoints have no guaranteed server-side ordering — any reorder will produce a phantom diff on every subsequent plan. Change to `schema.SetAttribute`.

coreweave/inference/resource_inference_gateway.go
```diff
@@ -0,0 +1,842 @@
...
+						MarkdownDescription: "Path-based routing configuration. Routes requests based on the URL path.",
+						Attributes:          map[string]schema.Attribute{},
+					},
+				},
+			},
+			"endpoint_configuration": schema.SingleNestedAttribute{
+				Optional:            true,
+				MarkdownDescription: "Additional endpoint configuration options.",
+				Attributes: map[string]schema.Attribute{
+					"additional_dns": schema.ListAttribute{
```
sosheskaz: DNS names are an unordered bag. If the API echoes `additional_dns` back in a different order than written, this `ListAttribute` produces spurious diffs. Change to `schema.SetAttribute`.

coreweave/inference/resource_inference_deployment.go
```diff
@@ -0,0 +1,837 @@
...
+						},
+					},
+					"priority": schema.Int64Attribute{
+						Optional:            true,
+						MarkdownDescription: "Priority for cross-deployment scaling (0–1000). Higher values win when there is contention.",
+						Validators: []validator.Int64{
+							int64validator.Between(0, 1000),
+						},
+					},
+					"capacity_classes": schema.ListAttribute{
```
sosheskaz, line 293: `ListAttribute` implies insertion order is semantically significant. The description doesn't document ordering semantics here, and a separate `priority` field handles cross-deployment precedence. If the API doesn't preserve insertion order, this will produce phantom diffs. Either switch to `schema.SetAttribute`, or document explicitly in `MarkdownDescription` that the order is load-bearing.

coreweave/inference/data_source_inference_deployment_parameters.go
```diff
@@ -0,0 +1,177 @@
...
+
+func (d *InferenceDeploymentParametersDataSource) Metadata(_ context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_inference_deployment_parameters"
+}
+
+func (d *InferenceDeploymentParametersDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		MarkdownDescription: "Retrieve available parameter values for CoreWeave Managed Inference deployments.",
+		Attributes: map[string]schema.Attribute{
+			"gateway_ids": schema.ListAttribute{
```
sosheskaz: The resource's `gateway_ids` is correctly `SetAttribute` — this data source exposes the same conceptual field as a `ListAttribute`, which is inconsistent. Change to `schema.SetAttribute`.

coreweave/inference/data_source_inference_deployment_parameters.go
```diff
@@ -0,0 +1,177 @@
...
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"allowed_keys": schema.ListAttribute{
+							Computed:    true,
+							ElementType: types.StringType,
+						},
+					},
+				},
+			},
+			"instance_types": schema.ListAttribute{
```
sosheskaz: `instance_types` is an API-enumerated bag where ordering is not significant. `ListAttribute` produces phantom diffs if the API returns them in a different order. Change to `schema.SetAttribute`.

coreweave/inference/data_source_inference_gateway_parameters.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (d *GatewayParametersDataSource) Metadata(_ context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_inference_gateway_parameters"
+}
+
+func (d *GatewayParametersDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		MarkdownDescription: "Retrieve available gateway parameters for CoreWeave Managed Inference.",
+		Attributes: map[string]schema.Attribute{
+			"zones": schema.ListAttribute{
```
sosheskaz: Available zone names are an unordered API enumeration. Change to `schema.SetAttribute`.

coreweave/inference/resource_inference_deployment.go
```diff
@@ -0,0 +1,837 @@
...
+
+	if status.GetCreatedAt() != nil {
+		m.CreatedAt = types.StringValue(status.GetCreatedAt().AsTime().Format(time.RFC3339))
+	}
+	if status.GetUpdatedAt() != nil {
+		m.UpdatedAt = types.StringValue(status.GetUpdatedAt().AsTime().Format(time.RFC3339))
+	}
+
+	// gateway_ids
+	gwVals := make([]attr.Value, 0, len(spec.GetGatewayIds()))
```
sosheskaz: Repo convention is `make([]attr.Value, len(src))` with index assignment (`vals[i] = ...`), not `make(..., 0, len(src))` + `append`. Same pattern in `setFromGateway` and `setFromCapacityClaim` in `resource_inference_gateway.go:689`.

docs/resources/inference_capacity_claim.md
```diff
@@ -0,0 +1,77 @@
...
+- `updated_at` (String) RFC3339 timestamp of when the capacity claim was last updated.
+
+<a id="nestedatt--resources"></a>
+### Nested Schema for `resources`
+
+Required:
+
+- `capacity_type` (String) The capacity type for the capacity claim. Must be one of: `CAPACITY_TYPE_SERVERLESS`, `CAPACITY_TYPE_CUSTOMER`.
+- `instance_count` (Number) The number of instances to reserve. Must be at least 1.
+- `instance_id` (String) The instance type to reserve by ID specifier (e.g. `gb200-4x`). Case insensitive.
```
sosheskaz: I'm guessing that this is passing through to the API, but it's suspect. Should it perhaps be `instance_type`? That's what we normally call these. `instance_id` implies the ID of an instance, not the ID of an instance type.

docs/data-sources/inference_deployment_parameters.md
```diff
@@ -0,0 +1,54 @@
...
+
+```terraform
+data "coreweave_inference_deployment_parameters" "deploy_params" {}
+
+output "gateway_ids" {
+  value = data.coreweave_inference_deployment_parameters.deploy_params.gateway_ids
+}
+
+output "instance_types" {
+  value = data.coreweave_inference_deployment_parameters.deploy_params.instance_types
```
sosheskaz, line 23: Doesn't this have a potential issue if there are >1 gateways with different zones in the account (I think we discussed this but wanted to double-check in Github)?

docs/data-sources/inference_capacity_claim_parameters.md
```diff
@@ -0,0 +1,31 @@
...
+### Read-Only
+
+- `zone_instance_types` (Attributes Map) Available instance types per zone (keyed by zone name). (see [below for nested schema](#nestedatt--zone_instance_types))
+
+<a id="nestedatt--zone_instance_types"></a>
+### Nested Schema for `zone_instance_types`
+
+Read-Only:
+
+- `instance_ids` (List of String)
```
sosheskaz: Should this be `ids`? So it's `instance_types[].id`?

coreweave/inference/resource_inference_deployment.go
```diff
@@ -0,0 +1,837 @@
...
+	f := deploymentFields{
+		Name:       m.Name.ValueString(),
+		GatewayIds: gwIds,
+		Disabled:   m.Disabled.ValueBool(),
+		Runtime: &inferencev1.DeploymentRuntime{
+			Engine: m.Runtime.Engine.ValueString(),
+		},
+		Resources: &inferencev1.DeploymentResources{
+			InstanceType: m.Resources.InstanceType.ValueString(),
+			GpuCount:     uint32(m.Resources.GpuCount.ValueInt64()), //nolint:gosec
```
sosheskaz, line 633: API note: not necessarily a provider issue, but this seems like the kind of thing that we might later realize should've been a list instead of a single field. I'm imagining this is somewhat analogous to an Auto Scaling Group, or another sort of managed instance group. This is especially relevant as we move into spot, where someone may want to serve inference out of spot capacity.

This can lead to nasty workarounds after-the-fact, for example, the [AWS Mixed Instance Policy](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/autoscaling_group#mixed-instances-policy-with-spot-instances-and-capacity-rebalance) implementation which has quite a few sharp edges, in large part due to its Launch Template specifically only supporting one instance type.

Is this an interface we want to commit to? Is a deployment which spans multiple instance types the kind of thing we might want to support in future? That may change the shape of what we want to do here.

kulkarnisamr: I don't think so, since spot == on-demand/serverless in this case and that gets handled through capacity claims for that org. all this is doing is saying I want this deployment on a `gd-8xh100ib-i128` and I don't think we will want to support a list of `instanceTypes` here. cc: @arsenetar

arsenetar: The spot/demand bit as @kulkarnisamr points out is technically handled elsewhere.  We considered a list here, but it introduces many problems at runtime as configurations for a deployment on particular hardware can vary drastically between different hardware.  Its much cleaner for users (troubleshooting etc) to not mix types.

coreweave/inference/resource_inference_deployment.go
arsenetar: The order is significant on this one.

docs/data-sources/inference_deployment_parameters.md
scotchneat: The way I understand it, the [CreateDeployments API](https://buf.build/coreweave/inference/docs/main%3Acoreweave.inference.v1alpha1#coreweave.inference.v1alpha1.CreateDeploymentRequest) operation accepts a list of gateway Ids. As long as one of them has a zone with the requested isntance_type, those gateways will handle the deployment. 

@arsenetar correct me if I'm wrong.

arsenetar: Yes as long as at least one gateway zone has the capability of handling the parameters then the deployment will be accepted.

coreweave/inference/resource_inference_capacity_claim.go
scotchneat: If I understand the protos, the capacity claims resources are mutable here, so I don't think `RequiresReplace` is appropriate here (verified with codex). 

`UpdateCapacityClaimRequest` carries the full `resources` block, and our provider maps all four Terraform fields (`instance_id`, `instance_count`, `capacity_type`, and `zones`) into that request via the shared capacity-claim resource builder. 
- coreweave/inference/resource_inference_capacity_claim.go:425: builds the shared proto payload for both create and update (maps all four Terraform fields into inferencev1.CapacityClaimResources)
- coreweave/inference/resource_inference_capacity_claim.go:465: `toUpdateCapacityClaimRequest` calls that same builder and assigns `Resources: f.Resources`

~I'm addressing the coverage gap, though. Added a provider-side test proving Terraform updates those resource fields through `UpdateCapacityClaim`.~

EDIT: It looks like Andres' commits already addressed this concern and applied the `RequiresReplace` suggestion.

coreweave/inference/resource_inference_deployment.go
scotchneat: The protos show these attributes are mutable, but the concern makes sense.
I could update the test to cover this case if the intended and expected behavior is that users can mutate them. Before I do, @kulkarnisamr can you confirm the API allows for this?

arsenetar: They are mutable.

sosheskaz (approved):

PR #293, opened 2026-03-27
feat(cks): add support for cks tailscale integration

ghost:
https://coreweave.atlassian.net/browse/KPLAT-1570

Adds support for accessing CKS clusters over Tailscale.


coreweave/cks/data_source_cluster.go
sosheskaz, line 1: Please change pr type/scope to `feat(cks)`

coreweave/cks/resource_cluster.go
```diff
@@ -911,6 +951,20 @@ func (r *ClusterResource) Schema(ctx context.Context, req resource.SchemaRequest
 					),
 				},
 			},
+			"tailscale": schema.SingleNestedAttribute{
+				Optional:            true,
+				MarkdownDescription: "Tailscale configuration for the cluster. Enables cluster access over a Tailscale VPN.",
```
sosheskaz: extreme nit

```suggestion
				MarkdownDescription: "Tailscale configuration for the cluster. Enables cluster access via a Tailscale VPN.",
```

coreweave/cks/resource_cluster.go
```diff
@@ -911,6 +951,20 @@ func (r *ClusterResource) Schema(ctx context.Context, req resource.SchemaRequest
...
+				Attributes: map[string]schema.Attribute{
+					"client_id": schema.StringAttribute{
+						Optional:            true,
```
sosheskaz, line 959: Should this be optional? The parent block being optional as well, implies an empty tailscale block is valid tailscale configuration.

```suggestion
						Required:            true,
```

cmtly: I wanted to be able to uninstall by either omitting the `tailscale` block or by doing this
```
  tailscale = {
    client_id = null
  }
```
This seemed to be needed to accomplish that.

coreweave/cks/resource_cluster_test.go
```diff
@@ -743,6 +750,101 @@ func TestClusterResource_V6FieldsCreateOnly(t *testing.T) {
...
+		InternalLBCidrNames: types.ListValueMust(types.StringType, []attr.Value{types.StringValue("internal-lb-cidr")}),
+	}
+
+	withTailscale := with(*base, func(c *cks.ClusterResourceModel) {
+		c.Tailscale = &cks.TailscaleResourceModel{
+			ClientID: types.StringValue("test-tailscale-client-id"),
+		}
+	})
+
+	ctx := context.Background()
```
sosheskaz:
```suggestion
	ctx := t.Context()
```
nit

coreweave/cks/resource_cluster.go
sosheskaz: Consider making `client_id` **Required** and letting `tailscale = null` (or omission) be the off-switch, rather than `tailscale = { client_id = null }`.

- `tailscale = null` is the idiomatic Terraform affordance for 'disabled / not configured.' Forcing users to write `tailscale = { client_id = null }` for the same intent is non-obvious and surfaces internal state shape into the user-facing config.
- Matches the rest of `ClusterResource`: `oidc`, `authn_webhook`, and `authz_webhook` all treat their inner fields as required-when-present.
- Eliminates the four-state collapse the collections nit flagged, and removes the `ClientID.IsNull()` preservation branch in `setTailscale` — the workaround exists only because the schema currently has to represent `tailscale = { client_id = null }` as a distinct state.

**One thing to confirm with the API:** does `UpdateCluster` with field-mask path `tailscale.client_id` and `tailscale` absent in the request actually clear server-side Tailscale state? If yes, the simpler design is strictly better. If the server needs to see the block explicitly to clear (e.g. mask path `tailscale` with a zero-value message), the current design has a reason to exist and should be called out in a comment.

wmcnamee-coreweave (dismissed):

scotchneat (dismissed):

scotchneat:
FYI the lab is currently experiencing capacity issues causing the clusters to never go ready. This results in the acctests failures we're seeing in the checks.

https://coreweave.slack.com/archives/C08KNRLEEDV/p1779833795628169

coreweave/cks/resource_cluster.go
```diff
@@ -54,6 +54,10 @@ type ClusterResource struct {
 	client *coreweave.Client
 }
 
+type TailscaleResourceModel struct {
+	ClientID types.String `tfsdk:"client_id"`
```
scotchneat, line 58: nit: I see the proto includes the `tailnet_domain` (output only) field. 
Was that intentionally left off from this schema because it's not necessary to maintain that in state?

Is there any case where the users would need the domain?

scotchneat (dismissed):
LGTM with one nit/question

coreweave/cks/resource_cluster.go
ChandonPierre: It takes 10-15 minutes into the initial provisioning process for the cluster to be created and connect to the tailnet and find out the FQDN. 

Since we can't return the value immediately or consistently it's better to leave it out, at least for now. 

It doesn't serve a functional purpose (clients already know their domain); it's more-so an indicator of "we've connected to your Tailnet, here's the proof"

scotchneat (approved):

PR #313, opened 2026-05-17
feat(object_storage): add acceptance sweeper and deprecate lifecycle rule.prefix

AndresRCB:
## Summary

Two changes to unblock CI on PR #279 (and any other branch that activates the `object_storage` acceptance suite):

1. **Add an acceptance test sweeper for `object_storage`** — registers two sweepers (buckets, org access policies), both guarded by the `tf-acc-objs-` prefix. CKS, networking, and inference all have this; object_storage had none, so leaked test buckets accumulated across runs and eventually collided with new tests' `tf-acc-objs-bucket-settings-13` name (the actual cause of the `TestBucketSettingsResource` "Already Exists" failures on recent runs).
2. **Deprecate `rule.prefix` on `coreweave_object_storage_bucket_lifecycle_configuration`** — CAIOS now rejects `PutBucketLifecycleConfiguration` requests that set the deprecated top-level S3 `Prefix` field (it was a silent no-op before, which could cause data loss when users assumed it scoped a lifecycle rule).

## Why these together

The PR #279 (inference) CI hit two distinct failures in `run-tests (object_storage) / test`, both unrelated to the inference work it adds:

- `TestBucketLifecycleConfiguration_MultiRule` failed because its fixtures set the deprecated `Prefix` field — fixed here.
- `TestBucketSettingsResource` failed because of a stranded bucket from an earlier run — the sweeper added here cleans that up before every CI acceptance run.

The matrix's `fail-fast: true` then cancelled the `inference`, `cks`, and `networking` jobs every time, so PR #279 never got a clean look. Landing this PR unblocks that.

## What's in each commit

- `97fcb4b` — `test(object_storage): add acceptance test sweeper` — new `coreweave/object_storage/sweeper_test.go` plus `.env` in `.gitignore`.
- `2f931a3` — `fix(object_storage): deprecate lifecycle rule.prefix; stop sending it` — schema `DeprecationMessage`, drop the field from `expandRules`, change `flattenLifecycleRules` to preserve client-side-only fields from the prior model so existing `prefix = "..."` configs don't churn, regenerate docs.
- `f912eb3` — `test(object_storage): move multi-rule fixtures to filter.prefix` — migrate the two test fixtures that were the last in-tree consumers of the deprecated field.

## Test plan

- [x] `make` (fmt + lint + install + generate) clean
- [x] `make test` passes
- [x] Local sweep against test org.
- [ ] CI acceptance for `object_storage` passes on this PR (verifies the sweeper inside the same matrix CI uses for downstream PRs)

## Backwards compatibility

- `rule.prefix` is *not* removed from the schema. Existing HCL keeps parsing, just emits a deprecation warning. State is preserved via the new `flattenLifecycleRules` preserve-by-ID logic, so users with the field set won't see phantom diffs on the next refresh.
- Users with `rule.prefix = "..."` and no `filter` block effectively had a no-op rule (CAIOS ignored the field). Behavior here is unchanged: we don't send anything, so the rule still applies to the whole bucket — same as it always did silently.

### DIFF
--- a/.github/workflows/acceptance-test.yaml
+++ b/.github/workflows/acceptance-test.yaml
@@ -21,6 +21,7 @@ jobs:
     env:
       COREWEAVE_API_ENDPOINT: ${{ inputs.suite == 'object_storage' && secrets.COREWEAVE_OBJS_API_ENDPOINT || vars.COREWEAVE_API_ENDPOINT }}
       COREWEAVE_API_TOKEN: ${{ inputs.suite == 'object_storage' && secrets.COREWEAVE_OBJS_API_TOKEN || secrets.COREWEAVE_API_TOKEN }}
+      TEST_ACC_SWEEP_ZONE: ${{ inputs.suite == 'object_storage' && 'US-EAST-04A' || 'US-LAB-01A' }}
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
       - uses: ./.github/actions/setup
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,5 @@ test/
 .vscode/*
 
 __debug*
+
+.env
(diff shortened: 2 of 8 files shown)


msillz (approved):
Approved!

PR #321, opened 2026-06-16
feat(inference): invalidate CAPACITY_TYPE_SERVERLESS, add superseding CAPACITY_TYPE_MANAGED

scotchneat:
Bumps inference proto to v1.22.0 - adds `CAPACITY_TYPE_MANAGED` and deprecates `CAPACITY_TYPE_SERVERLESS`.

## What changed

- `capacity_type` now accepts `CAPACITY_TYPE_MANAGED` (the replacement for `CAPACITY_TYPE_SERVERLESS`).
- `CAPACITY_TYPE_SERVERLESS` is deprecated: it is no longer advertised as a valid value.
- Valid values going forward: `CAPACITY_TYPE_CUSTOMER`, `CAPACITY_TYPE_MANAGED`.
- Acceptance Test configuration changes to support the above

## Impact on existing usage

- **Already-applied claims:** existing `CAPACITY_TYPE_SERVERLESS` claims were migrated to `CAPACITY_TYPE_MANAGED` **server-side**: no action needed for the resource to keep running. On the next `terraform refresh`/`plan`, state will reflect `CAPACITY_TYPE_MANAGED`.
- **Configurations still pinned to `CAPACITY_TYPE_SERVERLESS`:** `terraform plan`/`validate` will now **fail** with a validation error, e.g.:
  > `"CAPACITY_TYPE_SERVERLESS" is deprecated and is no longer an accepted value; use one of: ` `CAPACITY_TYPE_CUSTOMER`, `CAPACITY_TYPE_MANAGED`.
- **New/updated requests** using `CAPACITY_TYPE_SERVERLESS` are rejected by the API. The  provider surfaces this as a validation error at plan time.

## How to migrate

1. In your configuration, replace:
   ```hcl
   capacity_type = "CAPACITY_TYPE_SERVERLESS"
   ```
   with:
   ```hcl
   capacity_type = "CAPACITY_TYPE_MANAGED"
   ```
2. Run `terraform plan`. Because the server already migrated existing claims to `CAPACITY_TYPE_MANAGED`, the refreshed state matches your updated config, so no resource replacement is expected. The plan should show no changes for `capacity_type`.
3. `terraform apply` to clear the validation error and bring config/state back in sync.

> [!NOTE] 
> `capacity_type` forces replacement when changed. 
> If your state has not yet refreshed to `CAPACITY_TYPE_MANAGED` (still showing `CAPACITY_TYPE_SERVERLESS`) when you switch the config, the plan may propose a replacement. Running terraform refresh (or letting plan refresh) first ensures state reflects the server-side migration and avoids an unnecessary replace.


coreweave/deprecated_enum_validator.go
sosheskaz, line 1: We should move this to an `internal/` subpackage

coreweave/deprecated_enum_validator_test.go
```diff
@@ -0,0 +1,73 @@
...
+	t.Parallel()
+
+	tests := []struct {
+		name        string
+		value       types.String
+		wantWarning bool
+	}{
+		{
+			name:        "deprecated value warns",
+			value:       types.StringValue("CAPACITY_TYPE_SERVERLESS"),
```
sosheskaz: I like using string literals here because it guarantees contract stability. That may be good to note in a comment so someone doesn't see this later and use the enum strings as SoT.

However, if we do that, I'd prefer if we capture them as constants in the test package instead of using literals

docs/resources/inference_capacity_claim.md
```diff
@@ -49,7 +49,7 @@ resource "coreweave_inference_capacity_claim" "example" {
 
 Required:
 
-- `capacity_type` (String) The [capacity type](https://docs.coreweave.com/products/inference/concepts/scaling#capacity-claims) for the capacity claim. Must be one of: `CAPACITY_TYPE_SERVERLESS`, `CAPACITY_TYPE_CUSTOMER`.
+- `capacity_type` (String) The [capacity type](https://docs.coreweave.com/products/inference/concepts/scaling#capacity-claims) for the capacity claim. Must be one of: `CAPACITY_TYPE_SERVERLESS`, `CAPACITY_TYPE_CUSTOMER`, `CAPACITY_TYPE_MANAGED`. Note: `CAPACITY_TYPE_SERVERLESS` is deprecated and is rejected by the API for new or updated claims; existing claims are automatically migrated to `CAPACITY_TYPE_MANAGED`.
```
sosheskaz: `CAPACITY_TYPE_SERVERLESS` should be removed from the list here. That's a problem with the helper; override or extend the helper to support dropping fields from the primary listing

coreweave/inference/resource_inference_capacity_claim.go
```diff
@@ -162,10 +162,13 @@ func (r *InferenceCapacityClaimResource) Schema(_ context.Context, _ resource.Sc
...
 					"capacity_type": schema.StringAttribute{
 						Required:            true,
-						MarkdownDescription: fmt.Sprintf("The [capacity type](https://docs.coreweave.com/products/inference/concepts/scaling#capacity-claims) for the capacity claim. Must be one of: %s.", coreweave.EnumMarkdownValues(inferencev1.CapacityType_name, true)),
+						MarkdownDescription: fmt.Sprintf("The [capacity type](https://docs.coreweave.com/products/inference/concepts/scaling#capacity-claims) for the capacity claim. Must be one of: %s. Note: `CAPACITY_TYPE_SERVERLESS` is deprecated and is rejected by the API for new or updated claims; existing claims are automatically migrated to `CAPACITY_TYPE_MANAGED`.", coreweave.EnumMarkdownValues(inferencev1.CapacityType_name, true)),
 						PlanModifiers:       []planmodifier.String{stringplanmodifier.RequiresReplace()},
 						Validators: []validator.String{
 							stringvalidator.OneOf(coreweave.EnumValues(inferencev1.CapacityType_name, true)...),
+							// Warn (not reject) on deprecated values such as
+							// CAPACITY_TYPE_SERVERLESS; the API owns rejection.
+							coreweave.DeprecatedEnumValue(inferencev1.CapacityType(0).Descriptor()),
```
sosheskaz: I think we should be able to do something like
```suggestion
							coreweave.DeprecatedEnumValue(inferencev1.CapacityType_CAPACITY_TYPE_MANAGED.Descriptor()),
```
which reads a bit more clearly. (freehand, not checked)

coreweave/deprecated_enum_validator_test.go
```diff
@@ -0,0 +1,73 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			req := validator.StringRequest{
+				Path:        path.Root("capacity_type"),
+				ConfigValue: tt.value,
+			}
+			resp := &validator.StringResponse{}
+			v.ValidateString(context.Background(), req, resp)
```
sosheskaz: Per the repo convention (CLAUDE.md), tests should use `t.Context()` rather than `context.Background()` so the validator call honors test-lifecycle cancellation. Low practical impact here — the validator discards the context and `.golangci.yaml` does not enable `usetesting`, so this will not fail CI — but worth aligning for consistency.

```suggestion
			v.ValidateString(t.Context(), req, resp)
```

coreweave/inference/resource_inference_capacity_claim.go
```diff
@@ -162,10 +162,13 @@ func (r *InferenceCapacityClaimResource) Schema(_ context.Context, _ resource.Sc
 					},
 					"capacity_type": schema.StringAttribute{
 						Required:            true,
-						MarkdownDescription: fmt.Sprintf("The [capacity type](https://docs.coreweave.com/products/inference/concepts/scaling#capacity-claims) for the capacity claim. Must be one of: %s.", coreweave.EnumMarkdownValues(inferencev1.CapacityType_name, true)),
+						MarkdownDescription: fmt.Sprintf("The [capacity type](https://docs.coreweave.com/products/inference/concepts/scaling#capacity-claims) for the capacity claim. Must be one of: %s. Note: `CAPACITY_TYPE_SERVERLESS` is deprecated and is rejected by the API for new or updated claims; existing claims are automatically migrated to `CAPACITY_TYPE_MANAGED`.", coreweave.EnumMarkdownValues(inferencev1.CapacityType_name, true)),
```
sosheskaz: > The new description says existing claims are "automatically migrated to `CAPACITY_TYPE_MANAGED`", but `capacity_type` still carries `RequiresReplace()` on the next line. If the server migrates an existing `SERVERLESS` claim server-side, a refresh reads `MANAGED` into state while config still says `SERVERLESS` — that diff forces a destroy/recreate, and the recreate re-sends the now-rejected `SERVERLESS` value, leaving the user unable to apply without editing config.
> 
> Worth reconciling: either confirm `RequiresReplace` tolerates the `SERVERLESS`→`MANAGED` transition (e.g. a `RequiresReplaceIf` that ignores the deprecated→replacement migration), or soften the wording so users know they must update config to `CAPACITY_TYPE_MANAGED` rather than rely on transparent migration.

First pass: I think introducing a validator which marks `CAPACITY_TYPE_SERVERLESS` would be the right call. It's invalid config, and pushing a customer toward `RequiresReplace` if the replacement they create will be invalid, creates a case where the infrastructure will not converge toward the desired state when run.

coreweave/deprecated_enum_validator_test.go
scotchneat: Not sure why I missed this again. Thx

sosheskaz (approved):

PR #328, opened 2026-06-23
docs(networking_vpc): scope host_prefix /18 requirement to SUNK

dfinster:
## Summary

The `host_prefix` field on the `coreweave_networking_vpc` resource documented the `/18` CIDR mask as an **unconditional** requirement. This is misleading:

- The `/18` is a **SUNK-specific** recommendation (SUNK uses IP-based locality encoding for node rank ordering); non-SUNK VPCs have no such constraint.
- The provider does **not** enforce `/18` at runtime — a customer successfully created a VPC with `/22` host prefixes in `coreweave-us-west-01a`.
- Per-customer exceptions to `/18` already exist for SUNK customers (e.g. a `/21` granted on 2025-10-10).

Enterprise customers without large unused IPv4 blocks were reading the docs and believing they had to consume a `/18` (~16k addresses) per VPC even when a `/22` or smaller would suffice.

## Changes

All edits were made to the Go source (`coreweave/networking/resource_vpc.go`); the `docs/*.md` changes are generated via `make generate`.

- Reworded the deprecated `host_prefix` `MarkdownDescription` to scope the `/18` to SUNK clusters, note that exceptions may be granted per customer, and clarify that non-SUNK VPCs accept any prefix size.
- Added an equivalent SUNK note to the `host_prefixes` `MarkdownDescription`.
- Regenerated `docs/resources/networking_vpc.md`.

No behavioral/schema changes — documentation strings only.

### Note on the example migration

DOCS-2454 also requested migrating `examples/resources/coreweave_cks_cluster/resource.tf` off the deprecated `host_prefix`. That example already uses `host_prefixes` (done in #320), and a repo-wide grep confirms no remaining deprecated `host_prefix` usage in `examples/`, so no change was needed there.

## Test plan

- [x] `go build ./...` passes
- [x] `make generate` produces the regenerated docs included in this PR
- [ ] Confirm rendered Registry / docs.coreweave.com text reads correctly after merge

Resolves [DOCS-2454](https://coreweave.atlassian.net/browse/DOCS-2454)

Made with [Cursor](https://cursor.com)

[DOCS-2454]: https://coreweave.atlassian.net/browse/DOCS-2454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/coreweave/networking/resource_vpc.go
+++ b/coreweave/networking/resource_vpc.go
@@ -475,7 +475,7 @@ func (r *VpcResource) Schema(ctx context.Context, req resource.SchemaRequest, re
 			"host_prefix": schema.StringAttribute{
 				Optional:            true,
 				Computed:            true,
-				MarkdownDescription: "An IPv4 CIDR range used to allocate host addresses when booting compute into a VPC.\nThis CIDR must be have a mask size of /18. If left unspecified, a Zone-specific default value will be applied by the server.\nThis field is immutable once set.",
+				MarkdownDescription: "An IPv4 CIDR range used to allocate host addresses when booting compute into a VPC.\nFor SUNK clusters, this CIDR should have a mask size of /18 for optimal rank ordering; exceptions may be granted per customer. For non-SUNK VPCs, any prefix size is accepted. If left unspecified, a Zone-specific default value will be applied by the server.\nThis field is immutable once set.",
 				DeprecationMessage:  "`host_prefix` is deprecated. Use `host_prefixes` instead. The field will be removed in a future version. The equivalent expression for a given resource may be found by refreshing state and running `terraform state show coreweave_networking_vpc.example`.",
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplaceIfConfigured(),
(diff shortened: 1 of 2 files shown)


coreweave/networking/resource_vpc.go
```diff
@@ -475,15 +475,15 @@ func (r *VpcResource) Schema(ctx context.Context, req resource.SchemaRequest, re
 			"host_prefix": schema.StringAttribute{
 				Optional:            true,
 				Computed:            true,
-				MarkdownDescription: "An IPv4 CIDR range used to allocate host addresses when booting compute into a VPC.\nThis CIDR must be have a mask size of /18. If left unspecified, a Zone-specific default value will be applied by the server.\nThis field is immutable once set.",
+				MarkdownDescription: "An IPv4 CIDR range used to allocate host addresses when booting compute into a VPC.\nFor SUNK clusters, this CIDR should have a mask size of /18 for optimal rank ordering; exceptions may be granted per customer. For non-SUNK VPCs, any prefix size is accepted. If left unspecified, a Zone-specific default value will be applied by the server.\nThis field is immutable once set.",
```
sosheskaz, line 478: nit: "Rank ordering" is distributed-training jargon that the persona setting this field (infra eng) is relatively unlikely to know compared to a ML persona. This is likely google-able, and facilitates discovery of important information, with the prominent prescription readily available without overloading.

That said: this supports markdown. It might benefit from a link where a reader can learn more.

sosheskaz: > exceptions may be granted per customer

This raises the question: exceptions to what? "optimal rank ordering" to me implies that "deviating from this will use sub-optimal rank ordering", but the notion of "exceptions" may indicate that this is a feature which is on-or-off, depending on the CIDR mask.

sosheskaz (approved):
This might benefit from some persona-tailored helpers; non-blocking

PR #346, opened 2026-07-07
feat(cks): add kubelet configuration overrides to cks_cluster

brandonrjacobs:
## What

Adds a new optional `kubelet` attribute to the `coreweave_cks_cluster` resource (and the corresponding read-only attribute on the data source), exposing the CKS API's cluster-level kubelet override field.

This lets operators set kubelet options such as `maxPods`:

```hcl
resource "coreweave_cks_cluster" "example" {
  # ...
  kubelet = jsonencode({ maxPods = 256 })
}
```

## Why

The CKS API and proto (`Cluster`/`CreateClusterRequest`/`UpdateClusterRequest`) already carry a `kubelet` field (`google.protobuf.Struct`, "selective overrides applied to every cluster Node's kubelet configuration"), but the Terraform provider never surfaced it — so there was no way to set things like `maxPods` from Terraform. This closes that gap.

## Details

- **Type:** the attribute is a normalized-JSON string (`jsontypes.Normalized`). This is the natural fit for the arbitrary-JSON `structpb.Struct` field and, crucially, compares JSON **semantically** — so a value the API re-serializes with reordered keys does not produce a spurious diff (and therefore no unnecessary Node reboot, which the API requires on kubelet changes).
- **Validation:** a plan-time validator rejects non-object JSON (arrays/scalars) with a clear error, since the API expects an object.
- **Wiring:** threaded through create, update (adds the `kubelet` field-mask path, only when the value semantically changed), and read/flatten. Unset ⇄ null is preserved so an omitted block never drifts.
- **Data source:** the resource and data source share one model (`ClusterDataSourceModel = ClusterResourceModel`), so the field is also added (read-only) to the data source schema and its docs — required for the shared model to map cleanly.

## Testing

- `go build ./...`, `go vet ./...`, `gofmt` — all clean.
- `go test ./coreweave/cks/` — passes. Added unit coverage to `TestBuildUpdateRequest` (kubelet added / changed / removed / semantically-equal-noop) and a `TestClusterResource_Kubelet` acceptance test (create → add → reapply-noop → remove → import verify), mirroring the existing Tailscale test.
- Docs regenerated by hand to match `tfplugindocs` output (Terraform isn't available in my environment to run `make generate`); please re-run `make generate` in CI/locally to confirm no drift.

## Notes

- Adds one dependency: `github.com/hashicorp/terraform-plugin-framework-jsontypes` (official HashiCorp helper; the repo already uses the sibling `-nettypes`).

### DIFF
--- a/coreweave/cks/resource_cluster.go
+++ b/coreweave/cks/resource_cluster.go
@@ -13,6 +13,7 @@ import (
 	"github.com/coreweave/terraform-provider-coreweave/coreweave"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/hashicorp/hcl/v2/hclwrite"
+	"github.com/hashicorp/terraform-plugin-framework-jsontypes/jsontypes"
 	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/setvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
@@ -31,6 +32,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 	"github.com/zclconf/go-cty/cty"
 	"google.golang.org/protobuf/types/known/fieldmaskpb"
+	"google.golang.org/protobuf/types/known/structpb"
 )
 
 const (
@@ -164,6 +166,7 @@ type ClusterResourceModel struct {
 	SharedStorageClusterId      types.String              `tfsdk:"shared_storage_cluster_id"` //nolint:staticcheck
 	AdditionalServerSans        types.Set                 `tfsdk:"additional_server_sans"`
 	Tailscale                   *TailscaleResourceModel   `tfsdk:"tailscale"`
+	Kubelet                     jsontypes.Normalized      `tfsdk:"kubelet"`
 }
 
 func nodePortEmpty(np *cksv1beta1.PortRange) bool {
@@ -321,6 +324,7 @@ func (c *ClusterResourceModel) Set(cluster *cksv1beta1.Cluster) {
 	}
 
 	c.setTailscale(cluster)
+	c.setKubelet(cluster)
 
 	// Note: SharedStorageClusterId is not returned by the API, so we preserve it from the plan.
 	// This is intentional since it's marked as RequiresReplace - Terraform will manage this value.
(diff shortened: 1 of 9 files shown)


coreweave/cks/resource_cluster.go
```diff
@@ -504,6 +550,52 @@ func authWebhookChanged(plan, state *AuthWebhookResourceModel) bool {
...
+
+	s := &structpb.Struct{}
+	if err := s.UnmarshalJSON([]byte(req.ConfigValue.ValueString())); err != nil {
+		resp.Diagnostics.AddAttributeError(
+			req.Path,
+			"Invalid kubelet configuration",
+			fmt.Sprintf("The kubelet value must be a JSON object of kubelet configuration overrides: %s", err),
+		)
+	}
+}
```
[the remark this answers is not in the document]
scotchneat: Agreed.

coreweave/cks/update_mask_test.go
```diff
@@ -277,4 +279,57 @@ func TestBuildUpdateRequest(t *testing.T) {
...
+		state.Kubelet = jsontypes.NewNormalizedValue(`{"maxPods":256}`)
+		plan := baseModel()
+		// Same object, different key order/whitespace as the API might return.
+		plan.Kubelet = jsontypes.NewNormalizedValue(`{ "maxPods": 256 }`)
+		req := buildUpdateRequest(ctx, &plan, &state)
+		if req.Kubelet != nil {
+			t.Error("expected nil kubelet when semantically unchanged")
+		}
+		expectMaskPaths(t, req.UpdateMask.Paths)
+	})
```
scotchneat, line 336: Copilot suggests adding a plan time validator to reject an empty object (`{}`) for `kubelet`.

Once that's added we should also include a negative `"empty object is rejected"` test case to cover it.

```suggestion
	})

	t.Run("empty object is rejected", func(t *testing.T) {
		req := validator.StringRequest{Path: path.Root("kubelet"), ConfigValue: types.StringValue(`{}`)}
		resp := &validator.StringResponse{}
		kubeletValidator{}.ValidateString(ctx, req, resp)
		if !resp.Diagnostics.HasError() {
			t.Error("expected an error for an empty kubelet object, got none")
		}
	})
```

coreweave/cks/resource_cluster.go
```diff
@@ -504,6 +550,52 @@ func authWebhookChanged(plan, state *AuthWebhookResourceModel) bool {
...
+func kubeletChanged(ctx context.Context, plan, state *ClusterResourceModel) bool {
+	if plan.Kubelet.IsNull() != state.Kubelet.IsNull() {
+		return true
+	}
+	if plan.Kubelet.IsNull() {
+		return false
+	}
+	equal, _ := plan.Kubelet.StringSemanticEquals(ctx, state.Kubelet)
+	return !equal
+}
```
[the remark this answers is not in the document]
scotchneat: This seems like a false positive.

https://developer.hashicorp.com/terraform/plugin/framework/handling-data/terraform-concepts#unknown-values

> by the time a resource is expected to be created, read, updated, or deleted, only its computed attributes can be unknown. The rest are guaranteed to have known values (or be null).

coreweave/cks/resource_cluster.go
```diff
@@ -964,6 +1065,14 @@ func (r *ClusterResource) Schema(ctx context.Context, req resource.SchemaRequest
 					},
 				},
 			},
+			"kubelet": schema.StringAttribute{
+				CustomType:          jsontypes.NormalizedType{},
+				Optional:            true,
+				MarkdownDescription: "Selective overrides applied to every cluster Node's kubelet configuration, as a JSON object (e.g. `jsonencode({ maxPods = 256 })`). A Node reboot is required for changes to take effect, and unknown options are stored but ignored by CKS. See the [Kubernetes kubelet configuration reference](https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/) for supported options.",
+				Validators: []validator.String{
+					kubeletValidator{},
```
scotchneat, line 1082: nit: _not sure exactly where to anchor this_ — Claude+Codex found and independently verified an **_edge case_** worth flagging (not blocking this PR):

> Nothing stops a user from writing any field's value as a float, e.g. `{"memoryThrottlingFactor": 1.0}`. That's valid per `jsontypes.Normalized`. But the provider's own structpb.Struct handling (used to build the request) collapses that to `{"memoryThrottlingFactor": 1}` the moment it round-trips, since `structpb`'s number type is a plain `float64` with no memory of `1.0` vs `1`. Because `jsontypes.Normalized`'s semantic-equals compares JSON numbers as literal text rather than parsed value, config 1.0 vs. state 1 never reads as "unchanged" — it'll show a diff on every subsequent plan.

A plan modifier that compares the JSON structurally/numerically (rather than by literal text) between plan and state would suppress this.

scotchneat (changes requested):
Looks good. Requested one change and noted an edge case (API team will track).

coreweave/cks/update_mask_test.go
scott-cw: addressed in [7821fa1](https://github.com/coreweave/terraform-provider-coreweave/pull/346/commits/7821fa1f234cc8f6e70ffe825742fbc6e3829ef5)

scotchneat (approved):

PR #345, opened 2026-07-02
feat(object_storage): Add bucket inventory resource

nghosh1993:
## Summary

Adds a new Terraform resource, `coreweave_object_storage_bucket_inventory`, for managing S3 bucket inventory configurations on CoreWeave Object Storage (CAIOS). The resource supports the full lifecycle, namely, Create, Read, Update, Delete, and Import, and is registered with the provider.

**Following changes are added**:

New resource (resource_bucket_inventory.go)
- Full CRUD + ImportState 
- Schema covering name, enabled, included_object_versions, optional_fields (set), filter.prefix, schedule.frequency, and destination.bucket (bucket_arn, format, prefix, account_id), with validators on constrained fields (e.g. destination format).
- expandInventoryConfiguration / flattenInventoryConfiguration to convert between the Terraform model and the S3 API type.
- eqInventoryConfiguration to detect real drift and avoid spurious updates 
- waitForInventoryConfig to poll for eventual consistency after writes, since inventory config changes take a few seconds to propagate.
- Read treats a missing configuration as drift: it removes the resource from state so Terraform plans a recreate. Delete is idempotent on a not-found configuration/bucket.

Provider registration (provider.go)
- Registers NewBucketInventoryResource alongside the other object-storage resources.

## Tests
- `resource_bucket_inventory_unit_test.go`: schema shape, expand/flatten round-trips, and field validators (no network).
- `resource_bucket_inventory_test.go`: acceptance tests
  - TestBucketInventoryBasic — create → no-op plan → in-place update → no-op plan → import round-trip.
  - TestBucketInventoryDisappears — deletes the config out-of-band and asserts the Read path detects the disappearance and produces a non-empty (recreate) plan.

Notable fixes / details

- Destination bucket policy in acceptance tests. PutBucketInventoryConfiguration is rejected (IllegalInventoryConfigurationException) unless the inventory service can write to the destination bucket. The tests now attach a bucket policy granting that access and order the inventory resource after it via depends_on (the render helper gained an optional dependsOn parameter).

## Testing

```console
# Run against the QA object-storage environment 
export COREWEAVE_API_ENDPOINT=https://api.qa.coreweave.com                                         
export COREWEAVE_API_TOKEN=CW-SECRET-redacted
export COREWEAVE_S3_ENDPOINT=https://cwobject-staging.com 
TF_ACC=1 go test -v -run 'TestBucketInventoryBasic|TestBucketInventoryDisappears' ./coreweave/object_storage/
```

PR #352, opened 2026-07-13
ci(test): fail generate check on untracked generated files

scotchneat:
The original changes PR #345 hadn't included the generated docs changes and the `generate` check still succeeded:
https://github.com/coreweave/terraform-provider-coreweave/actions/runs/28622006850/job/84879970255?pr=345

Since `git diff --exit-code` only reports changes to tracked files, the generate job passed even when 'make generate' created brand-new docs pages that were never committed (e.g. PR #345's
docs/resources/object_storage_bucket_inventory.md). 

This registers untracked files under docs/ with git add -N (intent-to-add) first so new generated docs also fail the check, while keeping the compact-summary diff output.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,8 +37,9 @@ jobs:
       - run: make generate
       - name: git diff
         run: |
+          git add -N docs/ # make untracked generated docs visible to git diff
           git diff --compact-summary --exit-code || \
-            (echo; echo "Unexpected difference in directories after code generation. Run 'make generate' command and commit."; exit 1)
+            (echo; echo "Unexpected difference in directories after code generation (including new files). Run 'make generate' command and commit."; exit 1)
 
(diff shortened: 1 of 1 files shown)


sosheskaz (changes requested):
AI review: Requesting changes because this introduces a false negative for deleted generated docs.

`git add -N docs/` stages deletions of tracked files under `docs/`. The following unstaged `git diff --compact-summary --exit-code` therefore returns success when `make generate` deletes a tracked docs page. I reproduced this in an isolated repository: after the command, `git status --short` showed `D  docs/tracked.md`, the unstaged diff exited 0, and the cached diff exited 1.

Please preserve deletion detection while registering untracked files. The narrow change `git add -N --ignore-removal docs/` did so in the same test: new files remained visible to the unstaged diff, while deleted tracked files were not staged and also caused it to fail.

PR #356, opened 2026-07-23
fix(object_storage): isolate S3 client from AWS configuration

sosheskaz:
CoreWeave's S3 client loaded the AWS SDK's default external configuration even though the provider already supplies its endpoint, credentials, region, and HTTP transport. Ambient AWS settings could therefore alter client behavior or make construction fail when the SDK attempted to apply a custom CA bundle to the provider's retryable `*http.Client`.

## Changes

- Construct the S3 client directly from explicit CoreWeave-owned options.
- Preserve the existing endpoint, temporary credential cache, signing region, virtual-hosted addressing, HTTP timeout, retry policy, and transport.
- Remove the unused AWS default-config loader and its transitive dependencies.
- Reuse the normalized `s3Zone` during credential refresh, so empty-zone callers consistently use the existing `notempty` fallback.

## Test approach

The change was developed test-first with a standalone red test commit followed by the implementation:

- A default, isolated configuration constructed successfully before and after the change.
- `AWS_CA_BUNDLE` and shared-profile `ca_bundle` cases initially failed with the reported `WithTransportOptions` incompatibility.
- Malformed AWS environment configuration initially failed during default-config parsing.
- Conflicting shared-profile retry, FIPS, and dual-stack settings initially leaked into the CoreWeave client.
- All confirmatory cases pass after switching to explicit S3 options.

Positive validation:

- `go test ./coreweave/...`
- `go build ./...`
- `make generate`, with no tracked documentation changes
- Focused constructor tests under poisoned ambient AWS configuration
- Vulnerability scan produced no new findings relative to the pre-change baseline

`golangci-lint run` still reports the repository's existing `goconst` and `prealloc` findings; none are in the changed files.


wmcnamee-coreweave (dismissed):

coreweave/s3.go
```diff
@@ -61,18 +60,16 @@ func (c *Client) createS3Client(ctx context.Context, zone string) (*s3.Client, *
...
-		config.WithCredentialsProvider(credentials.NewStaticCredentialsProvider(resp.Msg.AccessKeyId, resp.Msg.SecretKey, "")),
-	)
-	if err != nil {
-		return nil, nil, err
-	}
-
-	s3Client := s3.NewFromConfig(awsConfig, func(o *s3.Options) {
-		o.UsePathStyle = false
-		o.BaseEndpoint = aws.String(c.s3Endpoint)
+	s3Client := s3.New(s3.Options{
```
AndresRCB, line 63: This change to direct `s3.Options` leaves `RequestChecksumCalculation` and `ResponseChecksumValidation` at unset. `LoadDefaultConfig` included them, so I think this changes behavior. I recommend setting them to what the default sets (`WhenSupported`) and adding them to the tests.

AndresRCB (changes requested):

coreweave/s3.go
sosheskaz: fixed in [0ee72b0](https://github.com/coreweave/terraform-provider-coreweave/commit/0ee72b042239ff7b6991045e29c12584a3de9a23)

AndresRCB (approved):

PR #359, opened 2026-07-23
fix(object_storage): use `Current` (not `Latest`) for bucket inventory included_object_versions

nghosh1993:
## What

The `coreweave_object_storage_bucket_inventory` resource validated and documented `included_object_versions` as accepting `All` and `Latest`. `Latest` is not a valid value, the S3 inventory API (and the underlying AWS SDK enum,
`InventoryIncludedObjectVersions`) only accepts `All` and `Current`. Any config using `Latest` would pass Terraform side validation and then be rejected by the backend.

This corrects the accepted value to `Current`. The CAIOS documentation that listed `Latest` was incorrect.

## Changes

- **Schema** (`resource_bucket_inventory.go`): `OneOf` validator and the attribute description now use `All` / `Current` instead of `All` / `Latest`.
- **Docs** (`docs/resources/object_storage_bucket_inventory.md`): regenerated description to match.
- **Unit test** (`resource_bucket_inventory_unit_test.go`): the validator table now asserts `Current` is accepted and `Latest` is rejected 
- **Acceptance test** (`resource_bucket_inventory_test.go`):
  - Updated `TestBucketInventoryBasic` to use `Current`.
  - Added `TestBucketInventoryIncludedObjectVersions`, which creates with `Current`, asserts a no-op re-apply, updates in place to `All`, and asserts a second no-op, verifying both valid values round-trip through the API and that switching between them is an in-place update (not a replace).

## Testing

- `go vet ./coreweave/object_storage/` passes.
- Unit tests pass.
- Acceptance tests (`TF_ACC=1`) require a live object-storage endpoint.

### DIFF
--- a/coreweave/object_storage/resource_bucket_inventory.go
+++ b/coreweave/object_storage/resource_bucket_inventory.go
@@ -114,8 +114,8 @@ func (r *BucketInventoryResource) Schema(ctx context.Context, req resource.Schem
 			},
 			"included_object_versions": schema.StringAttribute{
 				Required:            true,
-				MarkdownDescription: "Specifies which object versions are included in the inventory results. Valid values are `All` and `Latest`.",
-				Validators:          []validator.String{stringvalidator.OneOf("All", "Latest")},
+				MarkdownDescription: "Specifies which object versions are included in the inventory results. Valid values are `All` and `Current`.",
+				Validators:          []validator.String{stringvalidator.OneOf("All", "Current")},
 			},
 			"optional_fields": schema.SetAttribute{
 				Optional:            true,
--- a/coreweave/object_storage/resource_bucket_inventory.go
+++ b/coreweave/object_storage/resource_bucket_inventory.go
@@ -45,6 +45,17 @@ const (
 	ErrNoSuchInventoryConfiguration string = "NoSuchInventoryConfiguration"
 )
 
+// enumStrings converts an AWS SDK string-enum's Values() slice into the []string
+// form the OneOf validator expects, so the accepted values track the SDK enum
+// instead of being maintained as separate string literals.
+func enumStrings[T ~string](vals []T) []string {
+	out := make([]string, len(vals))
+	for i, v := range vals {
+		out[i] = string(v)
+	}
+	return out
+}
+
 // NewBucketInventoryResource returns a new resource instance.
 func NewBucketInventoryResource() resource.Resource {
 	return &BucketInventoryResource{}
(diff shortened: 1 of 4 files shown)


coreweave/object_storage/resource_bucket_inventory.go
```diff
@@ -114,8 +114,8 @@ func (r *BucketInventoryResource) Schema(ctx context.Context, req resource.Schem
 			},
 			"included_object_versions": schema.StringAttribute{
 				Required:            true,
-				MarkdownDescription: "Specifies which object versions are included in the inventory results. Valid values are `All` and `Latest`.",
-				Validators:          []validator.String{stringvalidator.OneOf("All", "Latest")},
+				MarkdownDescription: "Specifies which object versions are included in the inventory results. Valid values are `All` and `Current`.",
+				Validators:          []validator.String{stringvalidator.OneOf("All", "Current")},
```
sosheskaz: Where do these come from? Is it something we could get out of an AWS library or our object_storage protobufs somewhere?

coreweave/object_storage/resource_bucket_inventory_test.go
```diff
@@ -218,6 +218,99 @@ func TestBucketInventoryBasic(t *testing.T) {
 	})
 }
 
+// TestBucketInventoryIncludedObjectVersions exercises both valid
+// included_object_versions values — "Current" and "All" — proving each is
+// accepted by the API and that switching between them is planned and applied as
+// an in-place update (the attribute is not RequiresReplace).
+func TestBucketInventoryIncludedObjectVersions(t *testing.T) {
+	ctx := context.Background()
```
sosheskaz:
```suggestion
	ctx := t.Context()
```

coreweave/object_storage/resource_bucket_inventory_test.go
```diff
@@ -218,6 +218,99 @@ func TestBucketInventoryBasic(t *testing.T) {
...
+			name:         "no-op after update to All",
+			resourceName: resourceName,
+			bucket:       bucket,
+			inventory:    newInventory("All"),
+			configPlanChecks: resource.ConfigPlanChecks{
+				PreApply: []plancheck.PlanCheck{
+					plancheck.ExpectEmptyPlan(),
+				},
+			},
+		}),
```
sosheskaz, line 305: take or leave: might be nice to have a validation test-case in here that confirms `Latest` results in a failed plan. However, since the UTs exercise this, I don't think it's necessary.

coreweave/object_storage/resource_bucket_inventory.go
sosheskaz: I'd prefer if these were sourced from the defining library, which it sounds like is aws from the PR desc

nghosh1993: It comes from AWS library. Snippet from `aws-sdk-go-v2` `enums.go`  

Screenshot 2026-07-23 at 2 20 23 PM

coreweave/object_storage/resource_bucket_inventory_test.go
nghosh1993: good thought. More tests, the better :)

sosheskaz (approved):
approved with caveat: docs will go stale if more enum values are added

PR #366, opened 2026-07-24
feat(cks): expose tailscale tailnet_domain on cks_cluster

scott-cw:
## What

Exposes the Tailscale `tailnet_domain` (the server-assigned Tailnet DNS name) as a computed attribute on the `coreweave_cks_cluster` resource and data source, under the existing `tailscale` block. `client_id` was already wired; this surfaces the read-only domain the Tailscale operator assigns once the cluster is ready.

## Details

- `tailnet_domain` is `OUTPUT_ONLY` in the CKS API, so it's `Computed`. No `go.mod` bump needed - the pinned buf.build SDK already includes it.
- The operator populates it asynchronously, so `setTailscale` stores `null` (not `""`) until it's assigned, letting consumers distinguish "not yet assigned".
- No `UseStateForUnknown`: the value is async/mutable, so the plan stays "known after apply" on changes and every apply stays consistent. `tailscaleChanged` compares only `client_id`, so the async domain never triggers a spurious update.

## Tests

- Added `TestSetTailscale` unit test covering unset / empty-client-id / unassigned-domain (null) / assigned-domain cases.
- `go build`, `go vet`, and `go test ./coreweave/cks/` pass. Docs regenerated via `make generate`.
- Acceptance tests (`TF_ACC=1`) not run locally.

### DIFF
--- a/coreweave/cks/resource_cluster.go
+++ b/coreweave/cks/resource_cluster.go
@@ -57,7 +57,8 @@ type ClusterResource struct {
 }
 
 type TailscaleResourceModel struct {
-	ClientID types.String `tfsdk:"client_id"`
+	ClientID      types.String `tfsdk:"client_id"`
+	TailnetDomain types.String `tfsdk:"tailnet_domain"`
 }
 
 type AuthWebhookResourceModel struct {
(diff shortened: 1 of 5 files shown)


coreweave/cks/resource_cluster.go
```diff
@@ -1072,6 +1083,10 @@ func (r *ClusterResource) Schema(ctx context.Context, req resource.SchemaRequest
 							stringvalidator.RegexMatches(nonWhitespace, "must not be empty"),
 						},
 					},
+					"tailnet_domain": schema.StringAttribute{
+						Computed:            true,
+						MarkdownDescription: "The Tailscale Tailnet DNS name assigned to the cluster. Populated by the Tailscale operator after the cluster becomes ready.",
```
scotchneat, line 1088: This might be misleading. 

In [PR #293 discussion_r3494210437](https://github.com/coreweave/terraform-provider-coreweave/pull/293#discussion_r3494210437), @ChandonPierre explained that there's a 10-15 minute lag from when the cluster is provisioned to when we can know the FQDN. 

So an apply will complete when a cluster becomes ready (`STATUS_RUNNING`). The property will be null until that FQDN is known, which is usually not as soon as the cluster is ready.

Since this change handles the null, just rewording this should be enough. Something like: “May be null immediately after creation; populated on a subsequent refresh once the operator joins the tailnet.”

scotchneat (changes requested):
A few points to address or discuss:

* The `tailscale_domain` schema description wording concern (noted inline)
* nit: The change isn't covered in acceptance tests. Might be ok for this property considering the 1-15 minute delay, so up to you whether to add that.
* nit: It would be good to include the property in the examples -  examples/resources/coreweave_cks_cluster_tailscale/outputs.tf

PR #373, opened 2026-07-28
ci(mise): remove unused tools

sosheskaz:
Removes mise-managed tools that have no tracked repository consumers, including Prism. Go, `golangci-lint`, and Terraform remain pinned because the build, lint, and generation workflows invoke them.

## Why

The removed tools were installed by the common mise setup in every CI job without being executed. GoReleaser remains provided by `goreleaser-action`, which installs and invokes its own binary.

## Validation

- `make generate`
- `make test`
- Parsed the reduced config in isolation and resolved the three retained tools at their pinned versions
- `git diff --check`
- GitHub Actions passed build, snapshot build, generation, lint, and tests at `f374fb8`

## Caveats

A local clean-cache `make lint` reports 65 findings in unchanged Go files: 50 `goconst` and 15 `prealloc`. GitHub Actions lint passes at the same commit.

### DIFF
--- a/.config/mise.toml
+++ b/.config/mise.toml
@@ -3,19 +3,8 @@
 # https://github.com/jdx/mise/pull/6231
 [tools]
 go = "1.25.6"
-# make sure go is installed before go tools
-"go:golang.org/x/tools/cmd/godoc" = "v0.30.0"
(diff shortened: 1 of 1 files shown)


msillz (approved):
approved.

PR #381, opened 2026-07-30
feat(object_storage): add support for bucket archive settings

sam6258:


coreweave/object_storage/resource_bucket_settings.go
```diff
@@ -171,10 +199,50 @@ func (b *BucketSettingsResource) Schema(ctx context.Context, req resource.Schema
...
+		return
+	}
+
+	if data.ArchiveEnabled.ValueBool() && data.ArchiveAfterLastAccessDays.IsNull() {
+		resp.Diagnostics.AddAttributeError(
+			path.Root("archive_after_last_access_days"),
+			"Missing archive_after_last_access_days",
+			"archive_after_last_access_days must be set when archive_enabled is true.",
+		)
+	}
```
sosheskaz, line 267: **Blocking — this guards only one direction, and the unguarded direction is a permanent diff.**

It correctly rejects `archive_enabled = true` with no `archive_after_last_access_days`. It permits the reverse: a day count set while `archive_enabled` is `false`. The API drops the value in that case, so the config can never converge.

The proto is explicit about it — `cwobject/v1/cwobject.proto` on `archive_after_last_access_days`:

> Required when archive_enabled is true. **Ignored when archive_enabled is false.**

And CI confirms the consequence. [`run-tests (object_storage)`](https://github.com/coreweave/terraform-provider-coreweave/actions/runs/30592854115/job/91044488111) fails at **step 9 of 9**, the `"archive disabled"` step:

```
resource_bucket_settings_test.go:225: Step 9/9 error: After applying this test step,
the refresh plan was not empty.

  # coreweave_object_storage_bucket_settings.test_settings will be updated in-place
  ~ resource "coreweave_object_storage_bucket_settings" "test_settings" {
      + archive_after_last_access_days = 90
        # (3 unchanged attributes hidden)
    }
```

The `+` means state holds **null** while config wants `90`: the step applied `archive_enabled = false` with `archive_after_last_access_days = 90`, the server did not store the 90, `Read` refreshed it to null, and every later plan re-proposes it.

No state-handling change fixes this on its own. The config expresses something the API refuses to store, so it ought to be rejected at validation time. `resourcevalidator.Conflicting`/`AlsoRequires` can't express it either, since the trigger is a bool's *value* rather than its presence. Keeping it here in `ValidateConfig` is right, it just needs the second branch.

Note the second branch leans on `ValueBool()` folding a **null** `archive_enabled` into `false`. That is deliberate and correct — `ValidateConfig` sees the raw config, where `booldefault.StaticBool(false)` has not been applied yet, so an omitted toggle arrives as null and still means archive-off. It is worth knowing that this only holds while `archive_enabled` keeps its `false` default: without it, an omitted toggle would mean "leave the server's current archive state alone," and a retention value alongside it could be legitimate.

```suggestion
	if data.ArchiveEnabled.ValueBool() && data.ArchiveAfterLastAccessDays.IsNull() {
		resp.Diagnostics.AddAttributeError(
			path.Root("archive_after_last_access_days"),
			"Missing archive_after_last_access_days",
			"archive_after_last_access_days must be set when archive_enabled is true.",
		)
	}

	// A null archive_enabled means false here (that is the attribute's default),
	// so an omitted toggle must reject a retention value too.
	if !data.ArchiveEnabled.ValueBool() && !data.ArchiveAfterLastAccessDays.IsNull() {
		resp.Diagnostics.AddAttributeError(
			path.Root("archive_after_last_access_days"),
			"Unexpected archive_after_last_access_days",
			"archive_after_last_access_days can only be set when archive_enabled is true. The API ignores it while archive is disabled, which would leave a permanent diff.",
		)
	}
```

The `"archive disabled"` test step then needs its day value dropped — see the comment there.

---

**Why here rather than a declarative validator in the schema?** Worth recording, since it is the obvious next question. `boolvalidator` can *almost* express the forward half:

```go
boolvalidator.Any(
	boolvalidator.Equals(types.BoolValue(false)),
	boolvalidator.All(
		boolvalidator.Equals(types.BoolValue(true)),
		boolvalidator.AlsoRequires(path.MatchRoot("archive_after_last_access_days")),
	),
)
```

It is a trap in three ways, though. `AlsoRequires` fires on the attribute merely being non-null, so it cannot key on `true` without the `Equals` scaffolding. `Any` reports the union of its children's failures, so omitting the retention produces `Value must be "false"` alongside the real message. And all three stock validators early-return on null, so the omitted-toggle case — `archive_enabled` absent with a retention set — is validated by nothing at all, which is exactly the case that produces the drift, since the attribute defaults to false. It also only covers the forward direction; the reverse would need a second nested composition with `ConflictsWith`.

If you would rather have it schema-local, a custom `validator.Int32` on `archive_after_last_access_days` is the clean version. The framework calls attribute validators even when the config value is null (`internal/fwserver/attribute_validation.go` builds the request and loops unconditionally — the early-returns are the stdlib validators' own choice), so it can see the omitted case, read `archive_enabled` off `req.Config`, and attach both diagnostics to the correct path automatically. That is more machinery than this for a single pairing rule, so `ValidateConfig` seems right, but the option is real.

One thing that makes this check load-bearing rather than belt-and-braces: there is no `buf.validate` rule or `field_behavior` annotation on the field in `cwobject.proto`, so nothing enforces the pairing before the request leaves the provider.

coreweave/object_storage/resource_bucket_settings_test.go
```diff
@@ -119,6 +202,24 @@ func TestBucketSettingsResource(t *testing.T) {
...
+		// disabled state before the framework's destroy runs.
+		createBucketSettingsTestStep(ctx, t, bucketSettingsTestConfig{
+			name:         "archive disabled",
+			resourceName: resourceName,
+			bucket:       bucket,
+			settings: objectstorage.BucketSettingsResourceModel{
+				AuditLoggingEnabled: types.BoolValue(true),
+				ArchiveEnabled:      types.BoolValue(false),
+				// Retained in config while archive is off; the API ignores it.
+				ArchiveAfterLastAccessDays: types.Int32Value(90),
```
sosheskaz: **Blocking — this comment is the bug, and it is what fails CI.**

`// Retained in config while archive is off; the API ignores it.` — if the API ignores it, `Read` refreshes the attribute to null while the config still says `90`, so the plan never comes back empty. That is [step 9 of 9](https://github.com/coreweave/terraform-provider-coreweave/actions/runs/30592854115/job/91044488111):

```
Step 9/9 error: After applying this test step, the refresh plan was not empty.
  + archive_after_last_access_days = 90
```

`cwobject.proto` documents the same thing on the field itself — *"Ignored when archive_enabled is false."* Because this is deliberately the final applied step, the suite currently asserts that a permanently-drifting configuration is acceptable.

```suggestion
				ArchiveEnabled:      types.BoolValue(false),
```

Pairs with the `ValidateConfig` fix: once the reverse check exists, this combination is a config error, so leaving the day value here would fail for a different reason.

coreweave/object_storage/resource_bucket_settings_test.go
```diff
@@ -119,6 +202,24 @@ func TestBucketSettingsResource(t *testing.T) {
...
+			},
+			configPlanChecks: resource.ConfigPlanChecks{
+				PreApply: []plancheck.PlanCheck{
+					plancheck.ExpectResourceAction(fmt.Sprintf("coreweave_object_storage_bucket_settings.%s", resourceName), plancheck.ResourceActionUpdate),
+				},
+			},
+		}),
```
sosheskaz, line 235: **Follow-on, optional** — worth adding the regression test for the reverse validation so this drift class can't return silently.

It is an `ExpectError` step, so it fails during plan and never applies. That preserves the property the comment at L205-206 describes: `"archive disabled"` stays the last step that actually applies.

```suggestion
		}),
		{
			PreConfig: func() {
				t.Log("beginning coreweave_object_storage_bucket_settings archive days without enable test")
			},
			Config: objectstorage.MustRenderBucketResource(ctx, "test_bucket", &bucket) +
				objectstorage.MustRenderBucketSettingsResource(ctx, resourceName, &objectstorage.BucketSettingsResourceModel{
					Bucket:                     types.StringValue("coreweave_object_storage_bucket.test_bucket.name"),
					ArchiveEnabled:             types.BoolValue(false),
					ArchiveAfterLastAccessDays: types.Int32Value(90),
				}),
			ExpectError: regexp.MustCompile(`Unexpected archive_after_last_access_days`),
		},
```

The regex matches the diagnostic *summary* from the `ValidateConfig` suggestion rather than its detail text — Terraform wraps detail at ~78 columns, so a longer pattern can break across a line. Apply that suggestion too; on its own this step fails for lack of the error it expects.

coreweave/object_storage/resource_bucket_settings.go
```diff
@@ -98,7 +125,7 @@ func (b *BucketSettingsResource) Create(ctx context.Context, req resource.Create
 
 	_, err := b.client.SetBucketSettings(ctx, connect.NewRequest(&setReq))
 	if err != nil {
-		resp.Diagnostics.AddError("Error creating bucket settings", err.Error())
+		coreweave.HandleAPIError(ctx, err, &resp.Diagnostics)
 		return
 	}
```
sosheskaz: **The `SetBucketSettings` response is discarded, so state is written from the plan rather than from the server.**

`_, err := ...` throws away `SetBucketSettingsResponse.Settings`, and the `resp.State.Set(ctx, &data)` below persists the plan-derived model. `Read` and `ImportState` already do this correctly via `data.Set(getResp.Msg.Info.Settings)`.

This is why the drift in the `"archive disabled"` step surfaces at *refresh* rather than at apply: `Create` records `90` because it never asks the server what was actually stored. Reading the response back surfaces the disagreement at apply time, where it is far easier to diagnose.

I confirmed both halves of this against the bumped module before suggesting it — `SetBucketSettingsResponse` has a `settings` field (`cwobject/v1/cwobject.proto:553`), and `Set()` already early-returns on nil, so an empty response is safe.

```suggestion
	setResp, err := b.client.SetBucketSettings(ctx, connect.NewRequest(&setReq))
	if err != nil {
		coreweave.HandleAPIError(ctx, err, &resp.Diagnostics)
		return
	}

	data.Set(setResp.Msg.Settings)
```

Complement to the validation fix, not a substitute for it.

**One caveat before applying this.** `archive_after_last_access_days` is plain `Optional`, not `Computed`, so Terraform requires the post-apply value to equal the config value. That makes writing the response into state safe only if the server echoes back exactly what was requested. If it instead clamps a below-floor value to the org minimum, this trades today's refresh-drift for a hard `provider produced inconsistent result after apply`. The attribute description says the floor is "validated server-side", which reads like rejection rather than clamping — you'd know for certain. Note that making it `Computed` would not rescue this either — `Computed` only lets a *null* config resolve to whatever the server returns; a value the user actually set must still survive the round trip unchanged.

Applying this also invalidates the stated rationale in the L208-211 comment, which is premised on Create/Update copying the plan into state — see that thread.

coreweave/object_storage/resource_bucket_settings.go
```diff
@@ -189,7 +257,7 @@ func (b *BucketSettingsResource) Update(ctx context.Context, req resource.Update
 
 	_, err := b.client.SetBucketSettings(ctx, connect.NewRequest(&setReq))
 	if err != nil {
-		resp.Diagnostics.AddError("Error updating bucket settings", err.Error())
+		coreweave.HandleAPIError(ctx, err, &resp.Diagnostics)
 		return
 	}
```
sosheskaz: **Same as `Create`** — the response is discarded here too, so `Update` persists plan values rather than authoritative server state.

```suggestion
	setResp, err := b.client.SetBucketSettings(ctx, connect.NewRequest(&setReq))
	if err != nil {
		coreweave.HandleAPIError(ctx, err, &resp.Diagnostics)
		return
	}

	data.Set(setResp.Msg.Settings)
```

While you're in this file: the `HandleAPIError` migration is only half-done. `Create` (L128) and `Update` (L260) now use it, but `Read` (L174), `Delete` (L155), and `ImportState` (L276) still use raw `err.Error()`, so Connect error codes map inconsistently. Those lines fall outside this diff so I can't attach a suggestion to them, but the change is mechanical.

One caveat if you do it: `HandleAPIError` only *formats* the `CodeNotFound` diagnostic (`coreweave/client.go:70`) — it does not remove the resource from state. `Read` needs an explicit not-found branch too, or an out-of-band bucket deletion leaves a permanently un-reconcilable resource:

```go
if coreweave.IsNotFoundError(err) {
    resp.State.RemoveResource(ctx)
    return
}
```

Same `Optional`-vs-`Computed` caveat as on the `Create` comment applies here.

coreweave/object_storage/resource_bucket_settings.go
```diff
@@ -171,10 +199,50 @@ func (b *BucketSettingsResource) Schema(ctx context.Context, req resource.Schema
 				Computed:    true,
 				Default:     booldefault.StaticBool(false),
 			},
+			"archive_enabled": schema.BoolAttribute{
+				Description: "When true, idle STANDARD objects are archived to STANDARD_IA after `archive_after_last_access_days` without access. Your organization must be entitled to configure this setting.",
+				Optional:    true,
+				Computed:    true,
+				Default:     booldefault.StaticBool(false),
```
sosheskaz: **Nit, cosmetic** — the description text contains markdown (the backtick cross-reference), but it is assigned to `Description` rather than `MarkdownDescription`.

To be clear about what this does *not* do: it does not break the registry page. `terraform-plugin-docs` interpolates the `Description` fallback into the generated markdown either way, which is why the backticks render as inline code in this PR's `docs/` diff.

The mismatch is in what the schema advertises. The framework picks a description kind from whichever field you set — `MarkdownDescription` yields `markdown`, `Description` alone yields `plain` — so as written the schema declares plain text while carrying markdown. Consumers that honour that signal (`terraform providers schema -json`, editor hovers, third-party doc generators) render the backticks literally.

So: worth aligning for correctness of the declared kind, not to fix a visible rendering bug.

```suggestion
				MarkdownDescription: "When true, idle STANDARD objects are archived to STANDARD_IA after `archive_after_last_access_days` without access. Your organization must be entitled to configure this setting.",
				Optional:            true,
				Computed:            true,
				Default:             booldefault.StaticBool(false),
```

The suggestion spans four lines because renaming the key changes the longest key in the struct literal, so gofmt re-aligns the siblings — a one-line replacement would land unformatted and fail the `lint` job.

`archive_after_last_access_days` needs the same change at L213; I left it out because a suggestion can only replace a contiguous range and L208-211 sit in between. `make fmt` settles the alignment there.

Pre-existing precedent in this file — `audit_logging_enabled` uses plain `Description:` too — so entirely reasonable to defer.

coreweave/object_storage/resource_bucket_settings.go
```diff
@@ -171,10 +199,50 @@ func (b *BucketSettingsResource) Schema(ctx context.Context, req resource.Schema
...
+			},
+			// Deliberately not Computed: a Computed attribute with no default is
+			// planned as unknown when omitted from config, and Create/Update copy
+			// the plan straight into state. Leaving it Optional-only means omitting
+			// it yields a null the API request can skip entirely.
```
sosheskaz: Dropping `Computed` here is the right call, and documenting the reasoning is worth doing. But the rationale as written is premised on `Create`/`Update` copying the plan straight into state — which is the pre-existing behavior the comments on L128 and L260 suggest to change. Once those read the response back, this paragraph documents a constraint that no longer exists.

The decision still holds on independent grounds, which are worth capturing instead:

- With the `ValidateConfig` pairing, the server never originates a plan-time value — days is either user-supplied (archive on) or absent (archive off). `Computed` would advertise a server-filled value that never materialises.
- `Optional+Computed` would additionally need `UseStateForUnknown()`, because Terraform re-marks the null copied from prior state as unknown, so plans would otherwise show `(known after apply)` forever. That is two pieces of machinery to land where plain `Optional` already is.

```suggestion
			// Deliberately not Computed: given the ValidateConfig pairing with
			// archive_enabled, an omitted value always means "archive off", so it
			// should plan as a definite null, not an unknown for the server to fill.
```

For what it's worth, `Computed` does become *viable* once the response is read back — an unknown may legally resolve to null — so this is a "still the better choice" rather than "the only workable choice."

coreweave/object_storage/fake_cwobject_test.go
```diff
@@ -0,0 +1,294 @@
...
+)
+
+// fakeCWObject is an in-process stand-in for thoth's CWObject service, covering
+// the slice of its behavior that coreweave_object_storage_bucket_settings
+// depends on. It exists so the provider's handling of bucket settings can be
+// exercised without credentials, in particular for organizations that are not
+// entitled to bucket archive -- a population the acceptance suite cannot reach,
+// since it runs as a single org whose token is fixed by CI.
+//
+// Every rule below mirrors a specific behavior of the real service and cites
```
sosheskaz: These comments identify a non-public source repository and maps internal paths and implementation details, including data representation, feature gating, authorization, and persistence behavior. Please remove references that are non-public. This may be a hand-wave in some cases, but shouldn't name the repository or internal details for the benefit of readers.

coreweave/object_storage/resource_bucket_settings.go
```diff
@@ -171,10 +221,60 @@ func (b *BucketSettingsResource) Schema(ctx context.Context, req resource.Schema
...
+				Optional:            true,
+				Computed:            true,
+			},
+			// Deliberately not Computed: given the ValidateConfig pairing with
+			// archive_enabled, an omitted value always means archive is not being
+			// turned on, so it should plan as a definite null rather than an
+			// unknown for the server to fill.
+			"archive_after_last_access_days": schema.Int32Attribute{
+				MarkdownDescription: "Days since last access (or creation if never accessed) before a STANDARD object version is archived to STANDARD_IA. The default minimum is 60; your organization's entitlement may permit a different minimum, so the effective floor is validated server-side. Required when `archive_enabled` is true, and rejected otherwise.",
+				Optional:            true,
```
sosheskaz: **blocking:** `archive_after_last_access_days` also needs to accept server-originated state when archive configuration is omitted.

`SetBucketSettings` returns the stored settings object, and Create, Update, Read, and ImportState all copy it through `data.Set`. Consider an already-archived entitled bucket with remote state `archive_enabled = true` and `archive_after_last_access_days = 90`, plus an existing configuration that manages only `audit_logging_enabled`. `archive_enabled` may adopt `true` because it is Optional+Computed, but this attribute is planned as null; `data.Set` then returns `90`. On Create that violates Terraform plan/state consistency because only unknown planned values may change. On refresh or import, Terraform instead plans to remove `90`; Update omits the field, the service preserves and returns `90`, and the mismatch recurs.

Making this attribute Optional+Computed lets the pair absorb existing remote settings when both fields are omitted, while `ValidateConfig` still rejects explicit mismatched configuration. Please add regressions for an existing archive-enabled bucket under audit-only configuration and for import or removal of archive configuration while archive remains enabled.

coreweave/object_storage/fake_cwobject_test.go
dunnevan: Thanks, force pushed a new commit which removes them.

sosheskaz (approved):

PR #304, opened 2026-04-27
docs(object-storage): Scope bucket policy examples by org condition

thedriftofwords:
## Summary
- Updates Object Storage bucket policy Terraform examples to use one console user (`CW`) and one SAML group (`AWS`) as named principals, aligned with the merged docs PR [coreweave/docs#3044](https://github.com/coreweave/docs/pull/3044).
- Keeps `cw:PrincipalOrgID` in each example as an extra safety measure.
- Replaces the broad `arn:aws:s3:::*` data source example with bucket-specific bucket and object ARNs.
- Declares `coreweave/coreweave` as the provider source in each independently copyable example and regenerates the provider reference docs.

## Rationale
The earlier draft used `"CW": ["*"]`. In HCL, the square brackets are list syntax for the principal map, but `*` as the identifier is not the correct org-scoped wildcard form. The docs now use named principal ARNs for generic examples and reserve bare `"Principal": "*"` for explicit org-wide patterns with `cw:PrincipalOrgID`.

Each example is a standalone root module when copied from the generated reference documentation, so declaring `coreweave/coreweave` once per example makes it directly initializable without duplicating the declaration for every resource.

## Test plan
- [x] Run `terraform fmt` on the updated example files
- [x] Run `make -f GNUmakefile generate`
- [x] Run `go test ./coreweave/object_storage`
- [x] Run `git diff --check`
- [x] Run `terraform init -backend=false` and `terraform validate` for both updated examples.
- [x] Confirm `terraform plan -refresh=false` reaches provider configuration for both examples. It stops as expected without a CoreWeave API token; no remote resources are created.

## Additional validation with documentation example harness

To increase confidence before merge, the changes in this PR were also validated with a credential-free documentation example harness developed on branch `cg/DOCS-3002-terraform-example-harness` ([DOCS-3002](https://coreweave.atlassian.net/browse/DOCS-3002)). That branch is based on this PR and is not part of this merge scope.

The harness copies each registered example into a temporary directory so no `.terraform/` or lockfiles are written under `examples/`, then runs:

1. `terraform fmt -check`
2. `terraform init -backend=false -input=false`
3. `terraform validate`

Both modules in this PR passed:

- `examples/resources/coreweave_object_storage_bucket_policy`
- `examples/data-sources/coreweave_object_storage_bucket_policy_document`

[...]

PR #382, opened 2026-07-31
test(examples): Add credential-free documentation example harness

thedriftofwords:
## Summary
- Adds a credential-free harness that validates registered Terraform documentation examples with `terraform fmt -check`, `terraform init -backend=false`, and `terraform validate`.
- Exposes the command as `make -f GNUmakefile test-examples` and runs it in CI.
- Documents the command and registration workflow in `examples/README.md`.

## Dependency
**Draft until [coreweave/terraform-provider-coreweave#304](https://github.com/coreweave/terraform-provider-coreweave/pull/304) merges.**

This branch is rebased on PR #304 so the harness validates the updated bucket policy examples, including `required_providers` and named principal ARNs. After #304 lands, rebase this branch onto `main` and mark ready for review.

Related: [DOCS-3002](https://coreweave.atlassian.net/browse/DOCS-3002)

## Test plan
- [x] Run `make -f GNUmakefile test-examples`
- [x] Run `go test ./coreweave/object_storage`
- [x] Confirm no `.terraform/` or lockfiles are left under `examples/`
- [ ] CI `test-examples` job passes after rebase onto `main` post-#304

## Design notes
- Default harness is credential-free and stops at `validate` (no `plan` or `apply`).
- Examples are copied to temporary directories before validation so source trees stay clean.
- Register additional examples by adding paths to `tools/test-examples.sh`.

Made with [Cursor](https://cursor.com)

[DOCS-3002]: https://coreweave.atlassian.net/browse/DOCS-3002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -7,3 +7,17 @@ The document generation tool looks for files in the following locations by defau
 * **provider/provider.tf** example file for the provider index page
 * **data-sources/`full data source name`/data-source.tf** example file for the named data source page
 * **resources/`full resource name`/resource.tf** example file for the named data source page
+
+## Validating documentation examples
+
+Independently copyable examples should declare `coreweave/coreweave` in a `terraform.required_providers` block so Terraform resolves the correct provider when a user copies the example into a new root module.
+
+Run the credential-free validation harness from the repository root:
+
+```bash
+make -f GNUmakefile test-examples
+```
+
+The harness copies each registered example into a temporary directory, then runs `terraform fmt -check`, `terraform init -backend=false`, and `terraform validate`. It does not run `terraform plan` or `apply`, and it does not require CoreWeave API credentials.
+
+To register a new example, add its directory path to the `example_dirs` list in [tools/test-examples.sh](../tools/test-examples.sh).
--- a/tools/test-examples.sh
+++ b/tools/test-examples.sh
@@ -1,0 +1,73 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+log() {
+  echo "[test-examples] $*" >&2
+}
+
+fail() {
+  log "ERROR: $*"
+  exit 1
+}
+
+if ! command -v terraform >/dev/null 2>&1; then
+  fail "terraform is required but was not found in PATH"
+fi
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+repo_root="$(git -C "$script_dir" rev-parse --show-toplevel)"
+
+# Explicit registry of independently copyable documentation examples.
+example_dirs=(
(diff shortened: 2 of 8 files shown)


examples/README.md
```diff
@@ -7,3 +7,17 @@ The document generation tool looks for files in the following locations by defau
 * **provider/provider.tf** example file for the provider index page
 * **data-sources/`full data source name`/data-source.tf** example file for the named data source page
 * **resources/`full resource name`/resource.tf** example file for the named data source page
+
+## Validating documentation examples
+
+Independently copyable examples should declare `coreweave/coreweave` in a `terraform.required_providers` block so Terraform resolves the correct provider when a user copies the example into a new root module.
```
sosheskaz, line 13: Instead, could we have the `verifiable examples` configuration scaffold this config?

The `required_providers` block may be contextually useful, but if combined with other examples, it becomes invalid. Additionally, common convention for `required_providers` that customers will probably tend to use is to put it in its own file, like `providers.tf`. Frameworks like `terragrunt` can also pre-configure files like this in many configurations.

tools/test-examples.sh
```diff
@@ -0,0 +1,73 @@
...
+  tmpdir="$(mktemp -d)"
+  trap 'rm -rf "$tmpdir"' RETURN
+
+  log "Validating $example_rel"
+
+  cp "$example_dir"/*.tf "$tmpdir/"
+
+  terraform -chdir="$tmpdir" fmt -check -diff
+  terraform -chdir="$tmpdir" init -backend=false -input=false
+  terraform -chdir="$tmpdir" validate
```
sosheskaz, line 55: As written: if someone makes an update to the code, and updates the example at the same time, I believe this will fail.

Instead of this, I'd suggest using the `devtf` script or similar, which builds the provider from source and runs against that. I believe the `required_providers` is still necessary.

tools/test-examples.sh
```diff
@@ -0,0 +1,73 @@
...
+  fail "terraform is required but was not found in PATH"
+fi
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+repo_root="$(git -C "$script_dir" rev-parse --show-toplevel)"
+
+# Explicit registry of independently copyable documentation examples.
+example_dirs=(
+  "examples/resources/coreweave_object_storage_bucket_policy"
+  "examples/data-sources/coreweave_object_storage_bucket_policy_document"
```
sosheskaz, line 24: Is this temporary, or do we want to (eventually) run it against all examples?

PR #387, opened 2026-08-10
fix(inference): Stop advertising and accepting deprecated CAPACITY_TYPE_CUSTOMER

captainshar:
## What and why

`CAPACITY_TYPE_CUSTOMER` is deprecated and the API rejects it. [infr-api #775](https://github.com/coreweave/infr-api/pull/775) (merged 2026-07-21) dropped Customer CapacityClaim support: the value is marked `[deprecated = true]` in `coreweave/inference/v1alpha1/capacity_claim.proto`, with the generated comment *"No longer supported: kept only for wire compatibility; validation rejects it on create/update."* `CAPACITY_TYPE_MANAGED` is the only accepted value.

The provider still advertises and accepts it. Two things were going on:

1. **Stale protos.** `go.mod` pinned the inference modules at the `20260629172746` build — 2026-06-29, three weeks before the deprecation landed. So `EnumMarkdownValuesExcludingDeprecated` had nothing to filter and emitted both values.
2. **Two call sites using the non-deprecation-aware helpers.** The `capacity_type` `MarkdownDescription` already used the excluding variant, but the `OneOf` validator and the invalid-value diagnostic still read from `CapacityType_name`. `EnumValuesExcludingDeprecated`'s own doc comment calls out this exact case:

   > Use it to offer only currently-valid enum values (e.g. in a `stringvalidator.OneOf` list or generated docs) so deprecated values are neither advertised nor accepted.

Net effect today: `terraform plan` accepts `capacity_type = "CAPACITY_TYPE_CUSTOMER"` and the failure only surfaces at apply, when the API rejects it. The generated docs tell users it is valid.

## Changes

- Bump `buf.build/gen/go/coreweave/inference/{protocolbuffers,connectrpc}/go` to the `20260810160348` build, which carries the deprecation.
- `capacity_type` validator: `EnumValues(CapacityType_name, true)` → `EnumValuesExcludingDeprecated(...Descriptor())`, so the value fails at plan time.
- Invalid-value diagnostic: `EnumMarkdownValues` → `EnumMarkdownValuesExcludingDeprecated`, so the error stops listing a value the API rejects.
- Regenerate docs. The only rendered change is `capacity_type`, now `Must be one of: CAPACITY_TYPE_MANAGED.`

`CapacityType` is the only inference enum with a deprecated value today, so the `DeploymentAutoscaling_CapacityClass` and `BodyBasedRouting_APIType` validators are deliberately left alone.

## Testing

- `go build ./...`, `go vet ./coreweave/...` clean
- `go test ./coreweave/inference/...` passes
- `tfplugindocs generate` produces a one-line diff, no unrelated churn

## Related

- Docs side: [coreweave/docs#3416](https://github.com/coreweave/docs/pull/3416) removes `CAPACITY_TYPE_CUSTOMER` from the hand-authored Scaling page and the Inference OpenAPI overlay. `public-docs/platform/terraform/resources/inference_capacity_claim.mdx` is generated from this repo, so it stays wrong until this merges and the docs Terraform sync runs.
- Jira: [DOCS-2905](https://coreweave.atlassian.net/browse/DOCS-2905)

[DOCS-2905]: https://coreweave.atlassian.net/browse/DOCS-2905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/coreweave/inference/resource_inference_capacity_claim.go
+++ b/coreweave/inference/resource_inference_capacity_claim.go
@@ -169,7 +169,7 @@ func (r *InferenceCapacityClaimResource) Schema(_ context.Context, _ resource.Sc
 						MarkdownDescription: fmt.Sprintf("The [capacity type](https://docs.coreweave.com/products/inference/scaling#capacity-claims) for the capacity claim. Must be one of: %s.", coreweave.EnumMarkdownValuesExcludingDeprecated(inferencev1.CapacityType_CAPACITY_TYPE_MANAGED.Descriptor())),
 						PlanModifiers:       []planmodifier.String{stringplanmodifier.RequiresReplace()},
 						Validators: []validator.String{
-							stringvalidator.OneOf(coreweave.EnumValues(inferencev1.CapacityType_name, true)...),
+							stringvalidator.OneOf(coreweave.EnumValuesExcludingDeprecated(inferencev1.CapacityType_CAPACITY_TYPE_MANAGED.Descriptor())...),
 						},
 					},
 					"zones": schema.SetAttribute{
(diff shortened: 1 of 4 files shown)


captainshar:
Closing because I don't usually contribute to this repo - feel free to reopen if actually useful.

PR #389, opened 2026-08-11
feat(inference): Validate engine against API-advertised engines instead of a hardcoded list

JustinPerlman:
## What

Replaces the hardcoded `runtime.engine` allow-list (`OneOf("vllm", "dynamo-vllm")`) with plan-time validation against the engines the **API server** currently advertises. This unblocks `dynamo-sglang` (already registered server-side) and any future engine with no provider release.

## Why

The static `OneOf` was a second source of truth that drifts from the platform's runtime-engines config: every new engine required a provider change + release before it could be used. The API already exposes the authoritative engine set via `GetDeploymentParameters` → `RuntimeParameters.RuntimeVersions` (the same data the `coreweave_inference_deployment_parameters` data source surfaces), so the provider can validate against that instead.

Note: a Plugin Framework attribute `validator.String` runs offline and cannot call the API, so this is implemented in the resource's `ModifyPlan` hook (post-Configure, plan time) rather than as an attribute validator.

## How

- Add `ModifyPlan` on `InferenceDeploymentResource`: fetch `GetDeploymentParameters`, error if `runtime.engine` isn't among the advertised `RuntimeVersions` keys, with a `must be one of: …` message listing the live set.
- Guards: skip on destroy (null plan), skip when no client (offline `terraform validate`), skip unknown/null engine values, and defer to the server if the API advertises no engines.
- Remove the hardcoded `OneOf` validator; update the `engine` attribute description and regenerate the resource doc.

## Behavior change

Fully-offline `terraform validate` no longer flags an invalid engine — it is caught at `plan` instead (which already contacts the API). Invalid engines still fail before apply, now with the live list of valid engines.

## Note
Either this or https://github.com/coreweave/terraform-provider-coreweave/pull/390 needed to get Dynamo-SGLang in prod

### DIFF
--- a/coreweave/inference/resource_inference_deployment.go
+++ b/coreweave/inference/resource_inference_deployment.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
+	"sort"
+	"strings"
 	"time"
 
 	inferencev1 "buf.build/gen/go/coreweave/inference/protocolbuffers/go/coreweave/inference/v1alpha1"
@@ -48,6 +50,7 @@ const (
 var (
 	_ resource.Resource                = &InferenceDeploymentResource{}
 	_ resource.ResourceWithImportState = &InferenceDeploymentResource{}
+	_ resource.ResourceWithModifyPlan  = &InferenceDeploymentResource{}
 
 	hostnamePattern = regexp.MustCompile(`^[a-zA-Z0-9]([a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])?$`)
 	semverPattern   = regexp.MustCompile(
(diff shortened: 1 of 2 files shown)


coreweave/inference/resource_inference_deployment.go
```diff
@@ -410,6 +413,64 @@ func (r *InferenceDeploymentResource) Configure(_ context.Context, req resource.
...
 }
 
+// ModifyPlan validates runtime.engine against the engines the API server
+// currently advertises, instead of a hardcoded list that can drift from the
+// platform's runtime-engines source of truth. The available engines are the
+// keys of RuntimeParameters.RuntimeVersions returned by GetDeploymentParameters
+// — the same data the coreweave_inference_deployment_parameters data source
+// exposes. Adding an engine server-side thus makes it usable here with no
+// provider release.
+func (r *InferenceDeploymentResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
```
AndresRCB, line 420: Before we were just validating the input values in-place, but the logic got more involved now. Please add tests using a stub `GetDeploymentParameters` (or however you prefer that doesn't involve making an API call in a full E2E/Acceptance test) to cover an engine present only in the server response (e.g. `dynamo-sglang`), an absent engine with the sorted allowed-values diagnostic, and the intentional empty-response bypass. We can leave just the happy path in the acceptance test (which uses `vllm`).

AndresRCB (changes requested):

AndresRCB (approved):

PR #401, opened 2026-08-18
test(object_storage): migrate acceptance sweepers

sosheskaz:
Completes the Object Storage acceptance-sweeper migration on top of the shared runtime and CKS layer.

## Changes

- Migrate bucket and organization-policy sweepers to the shared bounded runtime.
- Preserve bucket emptying, organization scoping, ownership matching, idempotent deletion, and completion behavior.
- Validate the sweep selector before client setup so malformed or blank overrides cannot broaden cleanup.

Inference adoption is maintained in its own stacked PR above this layer.

## Validation

- `go test -race ./coreweave/object_storage`
- `git diff --check`

### DIFF
--- a/coreweave/object_storage/sweeper_test.go
+++ b/coreweave/object_storage/sweeper_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"os"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


coreweave/object_storage/sweeper_test.go
```diff
@@ -39,47 +122,30 @@ func init() {
 			if err != nil {
 				return fmt.Errorf("failed to build client: %w", err)
 			}
```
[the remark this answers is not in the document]
sosheskaz: Fixed in 674fc97. Object Storage now parses the sweep runtime immediately after zone validation and before creating the timeout context, applying environment defaults, or building the client. The shared inference registration follows the same order and passes the validated runtime into each resource-specific sweep configuration.

coreweave/object_storage/sweeper_test.go
```diff
@@ -20,17 +21,93 @@ import (
...
+		},
+		Name: func(info *cwobjectv1.BucketInfo) string { return info.GetName() },
+		Match: func(info *cwobjectv1.BucketInfo) bool {
+			return strings.HasPrefix(info.GetName(), AcceptanceTestPrefix) && info.GetLocation() == zone
+		},
+		Delete: func(ctx context.Context, info *cwobjectv1.BucketInfo) error {
+			if err := deleteBucket(ctx, client, info.GetName(), info.GetLocation()); err != nil {
+				return fmt.Errorf("failed to delete bucket %s: %w", info.GetName(), err)
+			}
+			return nil
```
AndresRCB, line 65: Nits on redundant error wrapping. Non-blocking:

>`testutil.Sweep` already wraps delete failures as `delete coreweave_object_storage_bucket "<name>": …`, and `deleteBucket` names the bucket in its own errors, so this wrapper repeats the name a third time. The VPC sweeper just returns the bare `delete VPC: %w`. Same applies to the org policy wrapper on line 90.

```suggestion
			return deleteBucket(ctx, client, info.GetName(), info.GetLocation())
```

coreweave/object_storage/sweeper_test.go
```diff
@@ -39,47 +116,28 @@ func init() {
...
-	// radius limited to acceptance-test artifacts.
-	resource.AddTestSweepers("coreweave_object_storage_organization_access_policy", &resource.Sweeper{
-		Name:         "coreweave_object_storage_organization_access_policy",
+func newOrgAccessPolicySweeper() *resource.Sweeper {
+	return &resource.Sweeper{
+		Name:         orgAccessPolicySweeperName,
 		Dependencies: []string{},
-		F: func(_ string) error {
+		F: func(region string) error {
+			if _, err := normalizeObjectStorageSweepRegion(region); err != nil {
```
AndresRCB: question: the org policy sweeper used to ignore the region entirely (`_ string`), and the comment in `init` still says it runs unconditionally on any zone sweep. Since the region never influences selection here, is rejecting a blank one intentional for symmetry with the bucket sweeper, or should this stay region-agnostic? If it stays, worth a one-line note so the next reader doesn't think the region scopes the sweep.

coreweave/object_storage/sweeper_test.go
```diff
@@ -88,38 +146,93 @@ func init() {
...
+		{name: "org policy/invalid runtime", callback: newOrgAccessPolicySweeper().F, selector: "zone-a", wantError: "parse TEST_ACC_SWEEP_PARALLEL as integer"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.ErrorContains(t, tt.callback(tt.selector), tt.wantError)
+		})
+	}
+
+	_, found := os.LookupEnv(provider.CoreweaveApiEndpointEnvVar)
+	require.False(t, found, "client setup must not run before region and runtime validation")
```
AndresRCB, line 208: Want to move this LookupEnv and assertion inside the `t.Run` above so we know what case caused the issue?

>This check runs once after all four subtests, so if one callback regresses and reaches `SetEnvDefaults`, the failure is reported at the end with no indication of which case leaked. Moving the `LookupEnv` assertion inside `t.Run` (as the networking sweeper test does) pinpoints the offending callback.

AndresRCB (approved):
Approving with comments because they're non-blocking.

PR #409, opened 2026-08-21
fix(object_storage): preserve bucket policy conditions

abatilo:
Bucket policy conditions were silently omitted because the Terraform
schema stores scalar strings while BuildPolicyDocument decoded values into
string slices.

Decode the schema-aligned values first, then normalize them for the JSON
policy model. Add a regression test for cw:PrincipalOrgID.

### DIFF
--- a/coreweave/object_storage/data_source_bucket_policy_document.go
+++ b/coreweave/object_storage/data_source_bucket_policy_document.go
@@ -123,6 +123,8 @@ func (sl StringList) MarshalJSON() ([]byte, error) {
 // Single‐string values get normalized into a slice of length 1.
 type Condition map[string]map[string][]string
 
+type scalarCondition map[string]map[string]string
+
 func (c *Condition) UnmarshalJSON(b []byte) error {
 	// First unmarshal into a generic map[string]json.RawMessage
 	var rawOps map[string]json.RawMessage
@@ -285,10 +287,13 @@ func BuildPolicyDocument(ctx context.Context, bm BucketPolicyDocumentModel) Poli
 		// condition
 		cond := Condition{}
 		if !sm.Condition.IsNull() {
-			var tmp Condition
+			var tmp scalarCondition
 			sm.Condition.ElementsAs(ctx, &tmp, false)
-			for op, raw := range tmp {
-				cond[op] = raw
+			for op, values := range tmp {
+				cond[op] = make(map[string][]string, len(values))
+				for key, value := range values {
+					cond[op][key] = []string{value}
+				}
 			}
 		}
 		// actions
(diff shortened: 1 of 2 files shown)


coreweave/object_storage/data_source_bucket_policy_document.go
```diff
@@ -285,10 +285,13 @@ func BuildPolicyDocument(ctx context.Context, bm BucketPolicyDocumentModel) Poli
 		// condition
 		cond := Condition{}
 		if !sm.Condition.IsNull() {
-			var tmp Condition
+			var tmp map[string]map[string]string
 			sm.Condition.ElementsAs(ctx, &tmp, false)
```
AndresRCB, line 291: We are discarding the diagnostics returned, so an unconvertible value removes the condition (i.e., we're swallowing decoding errors). That said, this is a pre-existing condition, so if you don't fix here, I can fix in another PR.

coreweave/object_storage/data_source_bucket_policy_document_unit_test.go
```diff
@@ -0,0 +1,33 @@
...
+				Condition: types.MapValueMust(types.MapType{ElemType: types.StringType}, map[string]attr.Value{
+					"StringEquals": types.MapValueMust(types.StringType, map[string]attr.Value{
+						"cw:PrincipalOrgID": types.StringValue("test-org-id"),
+					}),
+				}),
+			},
+		},
+	}
+
+	document := objectstorage.BuildPolicyDocument(context.Background(), model)
```
AndresRCB: nit: please use `t.Context()` here

AndresRCB (dismissed):

coreweave/object_storage/data_source_bucket_policy_document.go
```diff
@@ -285,10 +285,13 @@ func BuildPolicyDocument(ctx context.Context, bm BucketPolicyDocumentModel) Poli
 		// condition
 		cond := Condition{}
 		if !sm.Condition.IsNull() {
-			var tmp Condition
+			var tmp map[string]map[string]string
```
PleaseMeHaveNoRegrets: suggest to give this type a name. I recognize the underlying data type is changing, but a descriptive name here would be good practice for documentation

coreweave/object_storage/data_source_bucket_policy_document.go
PleaseMeHaveNoRegrets, line 291: the `Bucket policy conditions were silently omitted` is a problem of error handling in `ElementAs` where the `diagnostics` are provided as an output in `ElementAs` but not handled
https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework@v1.15.1/types/basetypes#MapValue.ElementsAs

PleaseMeHaveNoRegrets: suggestion to prevent silent omission would be to refactor `BuildPolicyDocument` - though i understand that may be a bigger lift than expected. 

I'll go ahead an create a task for this refactor - i would consider this out of scope for the PR

AndresRCB (approved):

PR #416, opened 2026-08-22
feat: add caller identity data source

abatilo:
Terraform configurations often need the organization identifier associated
with the configured API token. Users currently have to pass that value
separately or discover it outside Terraform.

Add `coreweave_caller_identity` so configurations can read the organization and
principal identifiers for the configured API token.

## Prior art and inspiration

Major HashiCorp cloud providers expose zero-argument, read-only data sources
for the identity or provider context associated with configured credentials:

- AWS [`aws_caller_identity`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/caller_identity)
  returns the account ID and calling entity ID in its `user_id` attribute.
  ([source](https://github.com/hashicorp/terraform-provider-aws/blob/38663e9229983da872026f8b6640ca88650fbc06/internal/service/sts/caller_identity_data_source.go#L22-L99))
- AzureRM [`azurerm_client_config`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/data-sources/client_config)
  returns the configured client, tenant, subscription, and object IDs.
  ([source](https://github.com/hashicorp/terraform-provider-azurerm/blob/69697b7fe6a923fc89e88f386bc95c3ed1ee2ec6/internal/services/authorization/client_config_data_source.go#L16-L60))
- Google [`google_client_openid_userinfo`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/client_openid_userinfo)
  returns the authenticated account email. It is the credential-identity
  analogue; `google_client_config` exposes project configuration rather than
  caller identity. ([source](https://github.com/hashicorp/terraform-provider-google/blob/e03b7622a26cf4e878f90bdccaca6aa3d03a7999/google/services/resourcemanager/data_source_google_client_openid_userinfo.go#L39-L138))

CoreWeave follows the same interface pattern with the two identifiers relevant
to its identity model: `organization_id` for the organization and
`principal_id` for the authenticated principal. The principal name supports
identities such as users and service accounts. The implementation uses
CoreWeave's own identity API and provider client conventions.

## Changes

- Add a read-only caller identity data source with `organization_id` and
  `principal_id` as its only attributes.
- Call `GET /v1beta1/auth/whoami` through the provider's configured endpoint,
  bearer token, user agent, HTTP timeout, and retry policy.
- Add protocol-level coverage and a dedicated read-only acceptance suite.
- Add the generated reference with links to the analogous AWS, AzureRM, and
  Google data sources, plus a minimal usage example.

## Validation

- `make generate`
- `make lint`
- `make test`
- `make build`
- `go mod tidy -diff`
- `git diff --check`
- `make testacc-sweep SUITES=caller_identity`
- `make testacc SUITES=caller_identity`


docs/data-sources/caller_identity.md
```diff
@@ -0,0 +1,28 @@
...
+data "coreweave_caller_identity" "current" {}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Read-Only
+
+- `organization_id` (String) The unique identifier of the organization associated with the configured CoreWeave API token.
+- `user_id` (String) The unique identifier of the user associated with the configured CoreWeave API token.
```
blairdrummond: Once we add service accounts, user_id might be a bit of a misnomer

In org Org Policies, Bucket policies, and maybe Access Policies(?) (need to double check) I think we use "Principal" to keep this generic. Maybe principal_id works?

abatilo: Good point. I renamed the unshipped attribute and matching model, tests, and documentation to `principal_id`. The endpoint response is already shaped as `principal.uid`, and the provider uses principal terminology for identities that can extend beyond users. The data source remains limited to `organization_id` and `principal_id`. Documentation generation, the full unit suite, lint, build, the suite sweep, and the live caller identity acceptance test pass on 8ed0e97.

wmcnamee-coreweave: Ya, I also support using `principal` here.

coreweave/client.go
```diff
@@ -18,7 +19,7 @@ import (
 	"google.golang.org/genproto/googleapis/rpc/errdetails"
 )
 
-func NewClient(endpoint string, s3Endpoint string, timeout time.Duration, interceptors ...connect.Interceptor) *Client {
+func NewClient(endpoint string, s3Endpoint string, timeout time.Duration, token string, userAgent string, interceptors ...connect.Interceptor) *Client {
```
AndresRCB: blocking: this changes the signature of the exported `NewClient` constructor, so existing callers that pass an interceptor as the fourth argument will no longer compile after upgrading. Please preserve the existing constructor through a compatible wrapper/new authenticated constructor, or explicitly mark and coordinate this as a breaking API change.

.github/workflows/acceptance-tests.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
         id: changes
         run: |
           # Available test suites - add new suites here
-          ALL_SUITES="cks networking object_storage inference"
+          ALL_SUITES="caller_identity cks networking object_storage inference"
```
AndresRCB: consider: the change detector maps this suite only to `coreweave/caller_identity/`, but the HTTP implementation lives in `coreweave/caller_identity.go`. A future PR changing only `GetCallerIdentity` would select no caller-identity acceptance job; please add that file to the suite mapping or colocate it under the suite directory.

coreweave/caller_identity/data_source_caller_identity_test.go
```diff
@@ -0,0 +1,146 @@
...
+const callerIdentityDataSourceType = "coreweave_caller_identity"
+
+type callerIdentityRequest struct {
+	method        string
+	path          string
+	authorization string
+	userAgent     string
+}
+
+func TestCallerIdentityDataSourceRead(t *testing.T) {
```
AndresRCB, line 26: There's happy path testting but no testing for malformed/missing identity or non-200 responses. Can you please cover possible errors with unexpected responses since your code does handle bad identity responses and also adds them to diagnostics?

Another motivation to ask for this is that the new logic for this data source (including some error handling) lives in  `coreweave/caller_identity.go` which is not included in the `acceptance-tests.yaml` scope for `caller_identity` (i.e., not part of the `calleridentity` package). That's not a big deal if these tests cover possible regressions caused by changing that code.

AndresRCB (changes requested):

coreweave/caller_identity/data_source_caller_identity_test.go
abatilo: Added protocol-level error coverage in 1831f4e. The table covers a 401 response, malformed JSON, a missing principal, a missing organization ID, and a missing principal ID. Each case reads through the Terraform protocol and asserts the `Unable to Read Caller Identity` error diagnostic, so it exercises both `coreweave/caller_identity.go` and the diagnostic boundary. The focused tests, full unit suite, lint, build, module tidy diff, and diff check pass.

AndresRCB (approved):

PR #417, opened 2026-08-25
feat(workload_federation): add OIDC config resource

hghafghazi:
## Summary

Adds the `coreweave_workload_federation_oidc_config` Terraform resource backed by the Workload Federation `WFControlPlaneService`.

- Supports create, read, update, delete, and import.
- Supports `name`, `description`, `issuer_url`, `audience`, and `active`.
- Exposes organization, activation, and timestamp metadata.
- Handles remote deletion and Terraform state drift.
- Preserves out-of-band deactivation when `active` is omitted.
- Allows HTTP issuer URLs only for loopback development issuers.
- Uses Unicode code-point length validation consistent with the API.
- Handles incomplete Create and Update responses without writing invalid state.
- Adds generated documentation and usage/import examples.
- Registers the resource and Workload Federation API client with the provider.

## Testing

- Added hermetic Terraform lifecycle tests using a local Connect server.
- Covered CRUD, import, remote deletion, activation changes, no-op updates, incomplete API responses, transient Update 404s, Unicode values, and issuer URL validation.
- Verified tests cannot be redirected by exported CoreWeave credentials.
- `go test ./...`
- `golangci-lint run`
- `make generate`
- `git diff --check`

## Jira

[CLDIAM-591](https://coreweave.atlassian.net/browse/CLDIAM-591)

[CLDIAM-591]: https://coreweave.atlassian.net/browse/CLDIAM-591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hghafghazi:
@claude /review-pr

examples/resources/coreweave_workload_federation_oidc_config/resource.tf
```diff
@@ -0,0 +1,6 @@
+resource "coreweave_workload_federation_oidc_config" "example" {
+  name        = "hcp-terraform"
+  description = "HCP Terraform workload identity"
+  issuer_url  = "https://app.terraform.io"
+  audience    = "coreweave.workload.identity"
```
blairdrummond: We should probably use `https://coreweave.com/iam` as the example audience

coreweave/workload_federation/resource_oidc_config.go
```diff
@@ -0,0 +1,459 @@
...
+	}
+	return types.StringValue(timestamp.AsTime().Format(time.RFC3339))
+}
+
+// SetFromProto replaces all remote fields in the model so out-of-band changes are reflected as drift.
+func (m *OIDCConfigResourceModel) SetFromProto(config *controlplanev1beta1.OIDCConfig) error {
+	if config == nil {
+		return fmt.Errorf("OIDC configuration is missing")
+	}
+	if config.GetUid() == "" {
```
AndresRCB: consider: A UID-only response is accepted here, after which `SetFromProto` writes empty strings into the required `name`, `issuer_url`, and `audience` attributes. That bypasses the incomplete-response recovery paths and can produce an inconsistent result after apply; could we reject responses missing those required fields and add a UID-only Create/Update case?

AndresRCB (dismissed):

examples/resources/coreweave_workload_federation_oidc_config/resource.tf
hghafghazi: fixed in https://github.com/coreweave/terraform-provider-coreweave/pull/417/commits/121991b132872568b7d281775a1feee007999f5a

AndresRCB (approved):

PR #422, opened 2026-08-27
feat(inference): Sync inference proto and enforce gateway name length

kulkarnisamr:
- Bump the inference buf modules to the latest snapshot to track current infr-api, and add the 38-character gateway name cap the API now enforces so it fails at plan time rather than on apply. Note capacity-claim instance type and zones are now case-sensitive.
- Expose gateway edge proxy mode and source IP allowlist

### DIFF
--- a/coreweave/inference/resource_inference_gateway.go
+++ b/coreweave/inference/resource_inference_gateway.go
@@ -267,7 +267,7 @@ func (r *InferenceGatewayResource) Schema(_ context.Context, _ resource.SchemaRe
 					},
 					"edge_proxy_mode": schema.StringAttribute{
 						Optional:            true,
-						MarkdownDescription: fmt.Sprintf("How client traffic reaches the gateway. Must be one of: %s. When unset, the platform default (currently `EDGE_PROXY_MODE_MANAGED`) applies.", coreweave.EnumMarkdownValues(inferencev1.EdgeProxyMode_name, true)),
+						MarkdownDescription: "How client traffic reaches the gateway. Must be one of: `EDGE_PROXY_MODE_MANAGED` (traffic is proxied through CoreWeave's managed edge network: DDoS mitigation and WAF, with TLS terminated at the edge and the origin address kept out of public DNS) or `EDGE_PROXY_MODE_DIRECT` (clients connect straight to the gateway's load balancer, with no edge filtering and the load balancer address published in DNS). When unset, the platform default (currently `EDGE_PROXY_MODE_MANAGED`) applies.",
 						Validators: []validator.String{
 							stringvalidator.OneOf(coreweave.EnumValues(inferencev1.EdgeProxyMode_name, true)...),
 						},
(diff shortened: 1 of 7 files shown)


coreweave/inference/resource_inference_gateway.go
```diff
@@ -844,6 +898,25 @@ func setFromGateway(m *InferenceGatewayResourceModel, gw *inferencev1.Gateway, p
...
+		if hasRanges {
+			rangeVals := make([]attr.Value, len(ec.GetAllowedSourceIpRanges()))
+			for i, r := range ec.GetAllowedSourceIpRanges() {
+				rangeVals[i] = cidrtypes.NewIPPrefixValue(r)
+			}
+			rangeSet, diags := types.SetValue(cidrtypes.IPPrefixType{}, rangeVals)
+			diagnostics.Append(diags...)
+			ecModel.AllowedSourceIPRanges = rangeSet
+		} else {
+			ecModel.AllowedSourceIPRanges = types.SetNull(cidrtypes.IPPrefixType{})
```
AndresRCB: You'll probably want to keep an empty set here instead of converting it to null because otherwise `allowed_source_ip_ranges = []` can produce state that differs from the plan and  Terraform could report an inconsistent result. I recommend an integration test to confirm this (just mocking the API) to cover this. (it's better if it's not an acceptance test).

coreweave/inference/resource_inference_gateway.go
```diff
@@ -259,6 +265,21 @@ func (r *InferenceGatewayResource) Schema(_ context.Context, _ resource.SchemaRe
 						Optional:            true,
 						MarkdownDescription: "Additional DNS names for the gateway endpoint. These DNS names must be manually configured to point to the gateway endpoint.",
 					},
+					"edge_proxy_mode": schema.StringAttribute{
+						Optional:            true,
+						MarkdownDescription: fmt.Sprintf("How client traffic reaches the gateway. Must be one of: %s. When unset, the platform default (currently `EDGE_PROXY_MODE_MANAGED`) applies.", coreweave.EnumMarkdownValues(inferencev1.EdgeProxyMode_name, true)),
```
AndresRCB: I recommend calling out what each mode means: `EDGE_PROXY_MODE_DIRECT` vs `EDGE_PROXY_MODE_MANAGED`. Not blocking for this provider PR (maybe you prefer that users go to your docs), but I would prefer to have it here if I were a Terraform user.

AndresRCB (changes requested):

coreweave/inference/resource_inference_gateway.go
kulkarnisamr: makes sense - switched to null-preservation via emptyOrNullSet: an explicit [] now round-trips as empty while unset stays null. Added two TestSetFromGateway_EndpointConfiguration subtests (mocking the API response, no acceptance test) covering both cases. Also applied the same fix to additional_dns, which had the identical latent issue.

coreweave/inference/resource_inference_gateway.go
kulkarnisamr: Done — the description now spells out MANAGED (proxied through the managed edge: DDoS/WAF, TLS at edge, origin kept out of DNS) vs DIRECT (clients hit the gateway LB directly, no edge filtering, LB address in DNS).

AndresRCB (approved):

PR #424, opened 2026-08-27
feat(auth): add refreshable access token source support

hghafghazi:
## Summary

Implements CLDIAM-592 by introducing an access-token source abstraction for CoreWeave API clients.

The provider continues to use static API tokens today, while the client can now support refreshable token sources required by future workload identity authentication.

## Changes

- Add the `AccessTokenSource` interface and static-token implementation.
- Resolve authorization tokens within the HTTP transport for every request attempt.
- Apply the same authentication path to:
  - ConnectRPC requests
  - Caller identity raw HTTP requests
  - CWObject access-key creation
- Refresh the token between retry attempts instead of replaying the original bearer token.
- Classify token-source failures into appropriate ConnectRPC error codes.
- Surface canceled, timed-out, unavailable, and unauthenticated failures through clear Terraform diagnostics.
- Prevent bearer tokens from being forwarded to a different origin.
- Warn when a request is intentionally sent without authentication due to an origin mismatch.
- Reject nil and typed-nil token sources during client construction.
- Correct transport error classification in the retry policy.

## Security

- Tokens are not included in state, diagnostics, or provider logs.
- Cross-origin requests do not receive the configured bearer token.
- Origin-mismatch warnings contain only sanitized origins, excluding URL paths and query parameters.

## Testing

- Static-token authentication
- Token refresh across raw HTTP and ConnectRPC retries
- Token-source error propagation and ConnectRPC classification
- Terraform diagnostic mapping
- Origin and redirect authentication behavior
- Typed-nil token-source rejection
- Sensitive query data exclusion from logs
- Retry behavior for transport and HTTP errors

Validation completed with:

- `go test ./...`
- `go test -race ./coreweave ./internal/auth ./internal/provider`
- `go vet ./...`
- `make lint`
- `git diff --check`

## Ticket

[CLDIAM-592](https://coreweave.atlassian.net/browse/CLDIAM-592)

[CLDIAM-592]: https://coreweave.atlassian.net/browse/CLDIAM-592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/coreweave/client.go
+++ b/coreweave/client.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"sync/atomic"
 	"time"
 
 	"buf.build/gen/go/coreweave/cks/connectrpc/go/coreweave/cks/v1beta1/cksv1beta1connect"
@@ -15,6 +16,7 @@ import (
 	"connectrpc.com/connect"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/coreweave/terraform-provider-coreweave/internal/auth"
 	retryablehttp "github.com/hashicorp/go-retryablehttp"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
@@ -25,19 +27,35 @@ type ClientOptions struct {
 	S3AttemptTimeout time.Duration
 }
 
-func NewClient(endpoint string, s3Endpoint string, timeout time.Duration, token string, userAgent string, interceptors ...connect.Interceptor) *Client {
-	return NewClientWithOptions(endpoint, s3Endpoint, timeout, token, userAgent, ClientOptions{}, interceptors...)
+var nextClientIdentity atomic.Uint64
+
+type cacheIdentityTokenSource interface {
+	CacheIdentity() string
+}
+
+func tokenSourceCacheIdentity(source AccessTokenSource) string {
+	if identified, ok := source.(cacheIdentityTokenSource); ok {
+		if identity := identified.CacheIdentity(); identity != "" {
+			return identity
+		}
+	}
+	return fmt.Sprintf("%T:%d", source, nextClientIdentity.Add(1))
+}
+
+// NewClient constructs a client that resolves a token for each HTTP attempt.
+func NewClient(endpoint string, s3Endpoint string, timeout time.Duration, tokenSource AccessTokenSource, userAgent string, interceptors ...connect.Interceptor) (*Client, error) {
+	return NewClientWithOptions(endpoint, s3Endpoint, timeout, tokenSource, userAgent, ClientOptions{}, interceptors...)
 }
 
 func NewClientWithOptions(
 	endpoint string,
 	s3Endpoint string,
 	timeout time.Duration,
-	token string,
+	tokenSource AccessTokenSource,
 	userAgent string,
 	options ClientOptions,
 	interceptors ...connect.Interceptor,
-) *Client {
+) (*Client, error) {
 	rc := retryablehttp.NewClient()
 	rc.HTTPClient.Timeout = timeout
 	rc.RetryMax = 10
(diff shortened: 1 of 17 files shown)


hghafghazi:
@claude /review-pr

coreweave/client.go
```diff
@@ -14,13 +14,15 @@ import (
...
+	"github.com/coreweave/terraform-provider-coreweave/internal/auth"
 	retryablehttp "github.com/hashicorp/go-retryablehttp"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"google.golang.org/genproto/googleapis/rpc/errdetails"
 )
 
-func NewClient(endpoint string, s3Endpoint string, timeout time.Duration, token string, userAgent string, interceptors ...connect.Interceptor) *Client {
+// NewClient constructs a client that resolves a token for each HTTP attempt.
+func NewClient(endpoint string, s3Endpoint string, timeout time.Duration, tokenSource AccessTokenSource, userAgent string, interceptors ...connect.Interceptor) (*Client, error) {
```
AndresRCB: consider: `NewClient` is exported, and this changes both its fourth argument and return arity, so any downstream caller stops compiling. Could the token-source path use a new constructor (or an option) while keeping the existing static-token constructor as a compatibility shim?

coreweave/client.go
```diff
@@ -73,7 +81,6 @@ type Client struct {
 	apiEndpoint string
 	httpClient  *http.Client
 	s3Endpoint  string
-	token       string
```
AndresRCB: Sorry, we had a conflicting urgent fix that affects this PR.

The new code has the S3 client caching keys by `c.token` to prevent credentials from one provider config being reused by another. This breaks that. You'll have to rebase and add an equivalent stable credential-source cache key (and a regression test across two auth principals), otherwise the process-wide cache can cross tenant/auth boundaries.

internal/auth/transport.go
```diff
@@ -0,0 +1,127 @@
...
+	}, nil
+}
+
+func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if !t.shouldAuthenticate(req.URL) {
+		tflog.Warn(req.Context(), "sending request without authentication: it is not addressed to the configured API endpoint", map[string]any{
+			"request_origin":    requestOrigin(req.URL),
+			"configured_origin": t.allowedScheme + "://" + t.allowedHost,
+		})
+		return t.base.RoundTrip(req)
```
AndresRCB: This mismatched-origin branch forwards the request unchanged, including the existing `Authorization` header (if present). That can happen for callers using the public interceptor hook, and Go preserves sensitive headers on some redirects (e.g., to subdomains), so the transport does not fully enforce the claimed no-cross-origin bearer guarantee. You'll have to do here what you do below: Clone the request and remove `Authorization` header, and extend the origin test to start with a bearer header to prove things work.

internal/provider/provider.go
```diff
@@ -192,17 +193,17 @@ func BuildClient(ctx context.Context, model CoreweaveProviderModel, tfVersion, p
...
 			return func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
-				req.Header().Add("Authorization", fmt.Sprintf("Bearer %s", token))
 				req.Header().Set("User-Agent", userAgent)
 				return next(ctx, req)
 			}
 		},
 	)
 
-	return coreweave.NewClient(endpoint, s3Endpoint, timeout, token, userAgent, headerInterceptor, coreweave.TFLogInterceptor()), nil
+	return coreweave.NewClient(endpoint, s3Endpoint, timeout, tokenSource, userAgent, userAgentInterceptor, coreweave.TFLogInterceptor())
```
AndresRCB: We extended `http_timeout` to S3 and this now conflicts with `resolveHTTPTimeouts`/`NewClientWithOptions`. You'll have to keep `S3AttemptTimeout` when wiring the token source during the rebase, or configured object-storage attempt timeouts will silently revert to the default (triggering a behavior we just fixed).

AndresRCB (changes requested):

AndresRCB (approved):

PR #426, opened 2026-08-28
fix(object_storage): bound bucket creation reconciliation

AndresRCB:
## Summary

- replace nested S3 transport retries with one AWS SDK retry layer capped at three attempts and use a single-send `CreateBucket`
- fail closed on incomplete owned-bucket inventory, reconcile ambiguous create responses only after exact ownership and zone proof, and persist bucket identity before bounded readiness/tag convergence
- share credential refresh safely across concurrent operations, including when the initiating caller is canceled, and apply positive `http_timeout` values consistently to CoreWeave API and S3 attempts
- add deterministic coverage for pagination, retry classification, one-shot mutation behavior, response-loss reconciliation, state retention, concurrency, and timeout handling

## Validation

- `go test ./... -count=1`
- `go test -race ./coreweave ./coreweave/object_storage ./internal/provider -count=1`
- `golangci-lint run`
- `make generate`
- repeated timeout and credential-refresh regression tests
- live Object Storage acceptance coverage for bucket lifecycle, tag updates/removal, replacement across names and zones, import/destroy, versioning, and lifecycle configuration using an internal organization

The live reproduction observed an immediate `HeadBucket` 400 during post-create routing convergence. The new readiness phase handles that known propagation response within the shared five-minute bound. An HTTP-attempt timeout while the Terraform operation context remains live is treated as ambiguous and reconciled without resending the mutation. Cancellation of the Terraform operation during the `CreateBucket` round trip remains unreconciled because follow-up ownership and zone checks cannot safely run on a dead operation context.

Configured `http_timeout` values now apply to both CoreWeave API and Object Storage S3 attempts; zero and negative values are rejected.

[CLDAPI-1063](https://coreweave.atlassian.net/browse/CLDAPI-1063)

[CLDAPI-1063]: https://coreweave.atlassian.net/browse/CLDAPI-1063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/coreweave/object_storage/resource_bucket.go
+++ b/coreweave/object_storage/resource_bucket.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	awsretry "github.com/aws/aws-sdk-go-v2/aws/retry"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/aws/smithy-go"
 	"github.com/aws/smithy-go/transport/http"
@@ -36,9 +37,10 @@ var (
 const (
 	errInvalidRegion = "InvalidRegion"
 
-	ErrNoSuchBucket string = "NoSuchBucket"
-	errNotFound     string = "NotFound"
-	errNoSuchTagSet string = "NoSuchTagSet"
+	ErrNoSuchBucket       string = "NoSuchBucket"
+	errNotFound           string = "NotFound"
+	errNoSuchTagSet       string = "NoSuchTagSet"
+	bucketReadMaxAttempts        = 8
 )
 
 func NewBucketResource() resource.Resource {
@@ -229,7 +231,10 @@ func (r bucketReadRetryer) GetAttemptToken(ctx context.Context) (func(error) err
 }
 
 func withBucketReadRetry(options *s3.Options) {
-	options.Retryer = bucketReadRetryer{Retryer: options.Retryer}
+	options.Retryer = awsretry.AddWithMaxAttempts(
+		bucketReadRetryer{Retryer: options.Retryer},
+		bucketReadMaxAttempts,
+	)
 }
 
 // waitForBucket polls HeadBucket every 'interval' until the bucket
--- a/coreweave/object_storage/resource_bucket.go
+++ b/coreweave/object_storage/resource_bucket.go
@@ -168,6 +168,21 @@ func isMissingBucketTagSetError(err error) bool {
 	return errors.As(err, &apiErr) && apiErr.ErrorCode() == errNoSuchTagSet
 }
 
+func bucketTagsForState(previous types.Map, remote []s3types.Tag) (types.Map, diag.Diagnostics) {
+	if len(remote) == 0 {
+		if previous.IsNull() {
+			return types.MapNull(types.StringType), nil
+		}
+		return types.MapValue(types.StringType, map[string]attr.Value{})
+	}
+
+	tagMap := make(map[string]attr.Value, len(remote))
+	for _, tag := range remote {
+		tagMap[*tag.Key] = types.StringValue(*tag.Value)
+	}
+	return types.MapValue(types.StringType, tagMap)
+}
+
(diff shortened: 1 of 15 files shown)


coreweave/object_storage/resource_bucket_create_unit_test.go
```diff
@@ -0,0 +1,727 @@
...
+}
+
+func immediatePhaseOptions(waits *int) s3PhaseOptions {
+	return s3PhaseOptions{
+		now:   time.Now,
+		delay: func(int) time.Duration { return 0 },
+		wait: func(ctx context.Context, _ time.Duration) error {
+			*waits++
+			return ctx.Err()
+		},
```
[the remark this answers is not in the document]
AndresRCB: Copilot is wrong. Invalid

coreweave/object_storage/resource_bucket.go
```diff
@@ -365,75 +330,37 @@ func (b *BucketResource) Create(ctx context.Context, req resource.CreateRequest,
...
 		return
 	}
 
-	if err := waitForBucket(ctx, s3Client, data.Name.ValueString(), true); err != nil {
-		handleS3Error(err, &resp.Diagnostics, data.Name.ValueString())
-		return
-	}
-
+	tags := []s3types.Tag(nil)
+	applyTags := !data.Tags.IsNull()
```
scotchneat: If i'm groking this right, this could change current behavior for configs with `tags = {}` (which is non-null). If so, this could result in a persistent null-versus-empty diff.

Essentially:

1. `tags = {}` == non-null. results in `applyTags == true`
2. `tags` stay nil because the empty map
3. `Create` calls `reconcileBucketAfterCreate(..., tags=nil, applyTags=true)` (though `expectedTags` is still nil)
4. The `runS3Phase` for `"bucket tag application"` would do an unnecessary `PutBucketTagging`. 

I'm not sure if that last part would fail.
Also, I don't see this specific scenario (`tags = {}`) covered in tests.

coreweave/object_storage/resource_bucket.go
```diff
@@ -448,23 +375,19 @@ func (b *BucketResource) Create(ctx context.Context, req resource.CreateRequest,
 				Value: aws.String(value),
 			})
 		}
+	}
```
scotchneat, line 397: I think switching up the `applyTags` logic and moving it to after the `tagMap` loop should handle the scenario.

```suggestion
	}
	
	applyTags := len(tags) > 0
```

coreweave/object_storage/resource_bucket.go
scotchneat:
```suggestion
```

scotchneat (changes requested):
I was going to approve with a nit, but after thinking it through, I want to draw attention to this.

Would prefer to maintain the current state behavior and cover the scenario in tests.

coreweave/object_storage/resource_bucket_test.go
```diff
@@ -90,21 +90,33 @@ func TestBucketResource(t *testing.T) {
...
-				TestName:     "initial bucket with tags",
+				TestName:     "initial bucket with explicit empty tags",
 				ResourceName: "test_acc_bucket",
 				Bucket: objectstorage.BucketResourceModel{
 					Name: types.StringValue(bucketName),
 					Zone: types.StringValue(zone),
-					Tags: types.MapValueMust(types.StringType, map[string]attr.Value{
-						"test-tag": types.StringValue("initial"),
-					}),
+					Tags: types.MapValueMust(types.StringType, map[string]attr.Value{}),
```
scotchneat: This diff removes the test I added in #420 to cover the post-create tags update issue. 

Can we retain the non-empty map coverage while adding the new empty-map coverage (might have to be another test)?

scotchneat (changes requested):

scotchneat (dismissed):

scotchneat (approved):

PR #427, opened 2026-08-28
feat(inference): Use ResourcesApplied condition for inference deployments create and update

kulkarnisamr:
Use ResourcesApplied condition for inference deployments create and update

### DIFF
--- a/coreweave/inference/resource_inference_deployment.go
+++ b/coreweave/inference/resource_inference_deployment.go
@@ -11,7 +11,6 @@ import (
 
 	inferencev1 "buf.build/gen/go/coreweave/inference/protocolbuffers/go/coreweave/inference/v1alpha1"
 	"connectrpc.com/connect"
-	"github.com/coreweave/terraform-provider-coreweave/coreweave"
 	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/setvalidator"
@@ -31,6 +30,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+
+	"github.com/coreweave/terraform-provider-coreweave/coreweave"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


coreweave/inference/resource_inference_deployment.go
```diff
@@ -76,6 +77,50 @@ var (
...
+		}
+		d := getResp.Msg.Deployment
+		status := d.GetStatus().GetStatus()
+		if status == inferencev1.Status_STATUS_ERROR || status == inferencev1.Status_STATUS_FAILED {
+			return d, status.String(), errDeploymentFailed
+		}
+		if resourcesApplied(d) {
+			return d, resourcesAppliedState, nil
+		}
+		return d, status.String(), nil
```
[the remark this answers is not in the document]
kulkarnisamr: `ResourcesApplied` condition is a precedent for `Ready` - it will always be there.

coreweave/inference/resource_inference_deployment_internal_test.go
```diff
@@ -97,3 +112,97 @@ func TestValidateEngineAvailable(t *testing.T) {
...
+	tests := map[string]struct {
+		deployment *inferencev1.Deployment
+		getErr     error
+		wantState  string
+		wantErr    error
+	}{
+		"resources applied while still creating completes": {
+			deployment: deploymentWith(
+				inferencev1.Status_STATUS_CREATING,
+				cond("ResourcesApplied", inferencev1.Condition_STATUS_TRUE),
```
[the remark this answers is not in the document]
kulkarnisamr: fixed.

coreweave/inference/resource_inference_deployment.go
```diff
@@ -76,6 +77,50 @@ var (
...
+	// wait on this synthetic state derived from the ResourcesApplied condition.
+	resourcesAppliedState = "RESOURCES_APPLIED"
+)
+
+// resourcesApplied reports whether the deployment's ResourcesApplied condition
+// is True — the completion signal for create/update. Readiness is left for the
+// user to poll via the API so Terraform runs don't block on model pulls/rollouts.
+func resourcesApplied(d *inferencev1.Deployment) bool {
+	for _, c := range d.GetStatus().GetConditions() {
+		if c.GetType() == conditionTypeResourcesApplied {
```
AndresRCB, line 93: This returns on the first matching condition, but deployments expose one `ResourcesApplied` condition per zone. Do you want to return "STATUS_TRUE" immediately after one zone shows `ResourceApplied`? If you do, you can just resolve this comment.

coreweave/inference/resource_inference_deployment.go
```diff
@@ -621,22 +652,8 @@ func (r *InferenceDeploymentResource) Update(ctx context.Context, req resource.U
...
-			}
-			d := getResp.Msg.Deployment
-			status := d.GetStatus().GetStatus()
-			if status == inferencev1.Status_STATUS_ERROR || status == inferencev1.Status_STATUS_FAILED {
-				return d, status.String(), errDeploymentFailed
-			}
-			return d, status.String(), nil
-		},
+		Target:     []string{resourcesAppliedState},
+		Refresh:    r.resourcesAppliedRefresh(ctx, deploymentID),
```
AndresRCB, line 656: `UPDATE` calls preserve conditions and change the spec, so a deployment can keep showing `True` during a new rollout, right? That would mean that the update can complete on its first poll using the previous `ResourcesApplied=True`. Is there a way to differentiate between the previous state and the new update's state?

coreweave/inference/resource_inference_deployment.go
kulkarnisamr: good question - there's two things here for single-zone/multi-zone:
1. single-zone - a zone where the referenced gateway isn't present never emits `ResourcesApplied`, so out-of-scope zones cannot short-circuit completion with a spurious `ResourcesApplied=True`.
2. multi-zone - a deployment referencing a multi-zone gateway is in-scope in several zones -> several `ResourcesApplied` conditions, and the first-True-wins loop returns as soon as one flips, before the others apply. currently, we only support single-zone gateways but when multi-zone gateways are supported, this will need to be fixed (creation of multi-zone gateways is gated by a feature flag which is disabled in prod).

coreweave/inference/resource_inference_deployment.go
kulkarnisamr: yes, I have https://github.com/coreweave/infr-api/pull/711 in progress which handles this differentiation. once that is merged, I will make the fixes here.

AndresRCB (approved):

PR #432, opened 2026-08-31
fix(object_storage): retry post-create InvalidRegion updates

scotchneat:
## Summary

Adds bounded retries for transient S3 `InvalidRegion` errors during bucket policy, versioning, lifecycle, and inventory Create/Update operations. This extends the tag-specific fix from [PR #420](https://github.com/coreweave/terraform-provider-coreweave/pull/420) and generalizes the convergence handling introduced in [PR #426](https://github.com/coreweave/terraform-provider-coreweave/pull/426).

## Findings

A successful `HeadBucket` does not guarantee that bucket subresource routes have converged. Customer logs showed accepted writes followed by `InvalidRegion` during readback, including resources managed separately from the bucket.

Customer context: [Slack thread](https://coreweave.slack.com/archives/C052XLUTY10/p1787905849768849?thread_ts=1787590757.398109&cid=C052XLUTY10)

The fix retries the affected write/readback sequence within one bounded timeout. Generic `400 BadRequest` retries remain limited to `HeadBucket`; global AWS retry behavior is unchanged.

The regression was added red in `8173e82` and made green by `ff079eb`.

Two notable call outs for reviewers:

- Mutation retries currently include `NoSuchBucket`/404, which could delay reporting a genuinely deleted bucket during Update.
- Direct wiring coverage could be strengthened for lifecycle, inventory, and versioning writes.

## Validation

- Focused post-create `InvalidRegion` regression
- Full object-storage test package
- `go build ./...`
- `go vet ./coreweave/...`

Tracking: [CLDAPI-1088](https://coreweave.atlassian.net/browse/CLDAPI-1088)

[CLDAPI-1088]: https://coreweave.atlassian.net/browse/CLDAPI-1088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/coreweave/object_storage/bucket_create.go
+++ b/coreweave/object_storage/bucket_create.go
@@ -2,11 +2,8 @@ package objectstorage
 
 import (
 	"context"
-	cryptorand "crypto/rand"
-	"crypto/tls"
 	"errors"
 	"fmt"
-	"math/big"
 	"net"
 	standardhttp "net/http"
 	"net/url"
@@ -18,18 +15,13 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	s3types "github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/aws/smithy-go"
-	smithyhttp "github.com/aws/smithy-go/transport/http"
-	"github.com/coreweave/terraform-provider-coreweave/coreweave"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 )
 
 const (
-	bucketPreflightPageSize  int32 = 10000
-	bucketPreflightMaxPages        = 10000
-	ambiguousCreateTimeout         = time.Minute
-	postCreateTimeout              = 5 * time.Minute
-	postCreateInitialBackoff       = 500 * time.Millisecond
-	postCreateMaxBackoff           = 5 * time.Second
+	bucketPreflightPageSize int32 = 10000
+	bucketPreflightMaxPages       = 10000
+	ambiguousCreateTimeout        = time.Minute
 )
 
 type bucketCreateAPI interface {
@@ -181,7 +173,7 @@ func reconcileAmbiguousCreate(
 	ctx, cancel := context.WithTimeout(parentCtx, ambiguousCreateTimeout)
 	defer cancel()
 
-	return runS3Phase(ctx, "ambiguous bucket create", bucket, zone, options, func(ctx context.Context) (bool, error) {
+	return runS3Phase(ctx, s3PhaseMetadata{phase: "ambiguous bucket create", bucket: bucket, zone: zone}, options, func(ctx context.Context) (bool, error) {
 		owned, err := bucketNameOwned(ctx, client, bucket)
 		if err != nil || !owned {
 			return false, err
@@ -199,82 +191,7 @@ func reconcileAmbiguousCreate(
 			return false, fmt.Errorf("owned bucket %q is in zone %q, want %q", bucket, actualZone, zone)
 		}
 		return true, nil
-	}, isPostCreateRetryableS3Error)
-}
-
-type s3PhaseOptions struct {
-	now   func() time.Time
-	delay func(attempt int) time.Duration
-	wait  func(context.Context, time.Duration) error
-}
-
-func (options s3PhaseOptions) withDefaults() s3PhaseOptions {
-	if options.now == nil {
-		options.now = time.Now
-	}
-	if options.delay == nil {
-		options.delay = postCreateBackoff
-	}
-	if options.wait == nil {
-		options.wait = waitForS3Backoff
-	}
-	return options
-}
-
-type s3PhaseError struct {
-	phase    string
-	attempts int
-	elapsed  time.Duration
-	err      error
-}
-
-func (e *s3PhaseError) Error() string {
-	return fmt.Sprintf("%s failed after %d attempt(s) in %s: %v", e.phase, e.attempts, e.elapsed.Round(time.Millisecond), e.err)
-}
-
-func (e *s3PhaseError) Unwrap() error {
-	return e.err
-}
-
-func runS3Phase(
-	ctx context.Context,
-	phase, bucket, zone string,
-	options s3PhaseOptions,
-	attempt func(context.Context) (bool, error),
-	retryable func(error) bool,
-) error {
-	options = options.withDefaults()
-	started := options.now()
-
-	for attemptNumber := 1; ; attemptNumber++ {
-		if err := ctx.Err(); err != nil {
-			return &s3PhaseError{phase: phase, attempts: attemptNumber - 1, elapsed: options.now().Sub(started), err: err}
-		}
-
-		done, err := attempt(ctx)
-		if err == nil && done {
-			return nil
-		}
-		if err != nil && !retryable(err) {
-			return &s3PhaseError{phase: phase, attempts: attemptNumber, elapsed: options.now().Sub(started), err: err}
-		}
-
-		delay := options.delay(attemptNumber)
-		errorClass, errorCode := s3ErrorMetadata(err)
-		tflog.Debug(ctx, "waiting for S3 operation convergence", map[string]interface{}{
-			"attempt":     attemptNumber,
-			"bucket":      bucket,
-			"elapsed":     options.now().Sub(started).String(),
-			"error_class": errorClass,
-			"error_code":  errorCode,
-			"next_delay":  delay.String(),
-			"phase":       phase,
-			"zone":        zone,
-		})
-		if waitErr := options.wait(ctx, delay); waitErr != nil {
-			return &s3PhaseError{phase: phase, attempts: attemptNumber, elapsed: options.now().Sub(started), err: waitErr}
-		}
-	}
+	}, isBucketPropagationRetryableS3Error)
 }
 
 func reconcileBucketAfterCreate(
@@ -285,10 +202,10 @@ func reconcileBucketAfterCreate(
 	applyTags bool,
 	options s3PhaseOptions,
 ) error {
-	ctx, cancel := context.WithTimeout(parentCtx, postCreateTimeout)
+	ctx, cancel := bucketPropagationContext(parentCtx)
 	defer cancel()
 
-	if err := runS3Phase(ctx, "bucket readiness", bucket, zone, options, func(ctx context.Context) (bool, error) {
+	if err := runS3Phase(ctx, s3PhaseMetadata{phase: "bucket readiness", bucket: bucket, zone: zone}, options, func(ctx context.Context) (bool, error) {
 		_, err := client.HeadBucket(ctx, &s3.HeadBucketInput{Bucket: aws.String(bucket)})
 		return err == nil, err
 	}, isBucketReadinessRetryableS3Error); err != nil {
@@ -303,16 +220,16 @@ func reconcileBucketAfterCreate(
 		Bucket:  aws.String(bucket),
 		Tagging: &s3types.Tagging{TagSet: expectedTags},
 	}
-	if err := runS3Phase(ctx, "bucket tag application", bucket, zone, options, func(ctx context.Context) (bool, error) {
+	if err := runS3Phase(ctx, s3PhaseMetadata{phase: "bucket tag application", bucket: bucket, zone: zone}, options, func(ctx context.Context) (bool, error) {
 		_, err := client.PutBucketTagging(ctx, tagInput)
 		return err == nil, err
-	}, isPostCreateRetryableS3Error); err != nil {
+	}, isBucketPropagationRetryableS3Error); err != nil {
 		return err
 	}
 
 	expected := append([]s3types.Tag(nil), expectedTags...)
 	slices.SortFunc(expected, cmpTag)
-	return runS3Phase(ctx, "bucket tag readback", bucket, zone, options, func(ctx context.Context) (bool, error) {
+	return runS3Phase(ctx, s3PhaseMetadata{phase: "bucket tag readback", bucket: bucket, zone: zone}, options, func(ctx context.Context) (bool, error) {
 		output, err := client.GetBucketTagging(ctx, &s3.GetBucketTaggingInput{Bucket: aws.String(bucket)})
 		if err != nil {
 			return false, err
@@ -323,7 +240,7 @@ func reconcileBucketAfterCreate(
 		actual := append([]s3types.Tag(nil), output.TagSet...)
 		slices.SortFunc(actual, cmpTag)
 		return slices.EqualFunc(expected, actual, eqTag), nil
-	}, isPostCreateRetryableS3Error)
+	}, isBucketPropagationRetryableS3Error)
 }
 
 func deleteBucketWithRetry(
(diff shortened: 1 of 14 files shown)


coreweave/object_storage/s3_convergence.go
```diff
@@ -0,0 +1,274 @@
...
+func runBucketMutationPhase(
+	ctx context.Context,
+	metadata s3PhaseMetadata,
+	options s3PhaseOptions,
+	mutate func(context.Context) error,
+) error {
+	return runS3Phase(ctx, metadata, options, func(ctx context.Context) (bool, error) {
+		err := mutate(ctx)
+		return err == nil, err
+	}, isBucketPropagationRetryableS3Error)
```
AndresRCB: This is the biggest concern IMO: We will end up retrying for 5 minutes on some operations where we should fail fast (like the one described below). That would be really bad and I recommend focusing on fixing that to avoid blocking the client.

>The mutation phase reuses the *propagation* classifier, so a `PutBucketPolicy`/`PutBucketVersioning`/etc. that returns 404 `NoSuchBucket` now spins for the full 5-minute budget instead of failing fast. That hits Create with a typo'd `bucket = "..."` (previously an immediate error), not just the deleted-bucket-on-Update case called out in the description. The ticket's DoD is scoped to *exact* `InvalidRegion`; the 404/`NoSuchTagSet` widening only makes sense for readbacks (a fresh config legitimately 404s until it propagates). Suggest two classifiers: mutation = `InvalidRegion` + the generic transient set (`isTransientBucketAPIError(code, false)` / 5xx / 429 / timeouts / network), readback = what you have today. If retrying `NoSuchBucket` on writes is intentional (literal bucket names racing bucket creation), it deserves its own, much shorter bound and a comment saying so.

coreweave/object_storage/s3_convergence.go
```diff
@@ -0,0 +1,274 @@
...
+func runBucketReadbackPhase(
+	ctx context.Context,
+	metadata s3PhaseMetadata,
+	options s3PhaseOptions,
+	converged func(context.Context) (bool, error),
+) error {
+	// The S3 control plane can return the just-written value from one replica
+	// before it is broadly visible. Preserve the pre-existing readback settle
+	// time so a single immediate response does not falsely signal convergence.
+	options.initialDelay = bucketPropagationReadbackDelay
```
AndresRCB: This overwrites whatever the caller put in `options.initialDelay`, so the new field is dead for callers and now every Create and Update of policy/versioning/lifecycle/inventory takes 5s sleep before the first read. I get that it preserves the old `PollUntil` ticker semantics, but a delay won't guarantee convergence and the mismatch loop covers that anyway. I would prefer to read immediately or make it a per-call option instead of a default override inside the generic helper.

coreweave/object_storage/s3_convergence.go
```diff
@@ -0,0 +1,274 @@
...
+			"error_code":  errorCode,
+			"next_delay":  delay.String(),
+			"phase":       metadata.phase,
+		}
+		if metadata.zone != "" {
+			logFields["zone"] = metadata.zone
+		}
+		tflog.Debug(ctx, "waiting for S3 operation convergence", logFields)
+		if waitErr := options.wait(ctx, delay); waitErr != nil {
+			return &s3PhaseError{phase: metadata.phase, attempts: attemptNumber, elapsed: options.now().Sub(started), err: waitErr}
```
AndresRCB, line 123: I.e., we're hiding the S3 error here now.

>When the wait returns `DeadlineExceeded`, the error from the *last attempt* is dropped, so after five minutes of `InvalidRegion` the user sees `bucket policy readback failed after 60 attempt(s) in 5m0s: context deadline exceeded` with no hint that it was `InvalidRegion` the whole time. The ticket DoD asks for an actionable diagnostic on persistent `InvalidRegion`. Carry the last attempt error too (a `lastErr` field, or `errors.Join(waitErr, err)`) so `errors.Is(..., DeadlineExceeded)` still works but the diagnostic names the S3 error. Related: because `s3PhaseError` wraps an `OperationError`, `handleS3Error`'s `errors.As` matches the inner error and the phase/attempts/elapsed prefix never reaches the diagnostic either — worth a case for `*s3PhaseError` there.

coreweave/object_storage/resource_bucket_policy.go
```diff
@@ -36,14 +35,21 @@ func NewBucketPolicyResource() resource.Resource {
 }
 
 type BucketPolicyResource struct {
-	client *coreweave.Client
+	client                   *coreweave.Client
+	s3ClientForConvergence   func(context.Context) (bucketPolicyAPI, error)
+	bucketPropagationOptions s3PhaseOptions
```
AndresRCB, line 40: I liked this recommendation a lot when exploring options with AI (the embedded struct recommended below fits the setup well):

>These test seams exist only on the policy resource; versioning, lifecycle and inventory pass `s3PhaseOptions{}` literals and take the concrete `*s3.Client`, so their Create/Update wiring can't be exercised the way `TestBucketPolicyCreateRetainsStateWhenReadbackTimesOut` does (you flagged this gap in the description). `BucketResource` already has the same shape (`postCreateOptions` + `getS3Client`). Pick one pattern and apply it to all four — a small embedded struct like `bucketSubresourceConvergence{ client func(ctx) (T, error); options s3PhaseOptions }` would keep the per-resource boilerplate to a couple of lines and let one table-driven test cover state retention on readback timeout for every subresource.

coreweave/object_storage/bucket_create.go
```diff
@@ -285,10 +202,10 @@ func reconcileBucketAfterCreate(
 	applyTags bool,
 	options s3PhaseOptions,
 ) error {
-	ctx, cancel := context.WithTimeout(parentCtx, postCreateTimeout)
+	ctx, cancel := bucketPropagationContext(parentCtx)
 	defer cancel()
 
-	if err := runS3Phase(ctx, "bucket readiness", bucket, zone, options, func(ctx context.Context) (bool, error) {
+	if err := runS3Phase(ctx, s3PhaseMetadata{phase: "bucket readiness", bucket: bucket, zone: zone}, options, func(ctx context.Context) (bool, error) {
```
AndresRCB, line 208: Are you thinking about improving this readiness check as well (a 200 HeadBucket below doesn't mean things are good)? For instance treating `InvalidRegion` from a subresource (like `GetBucketVersioning`) as "not ready"? What you're doing in the rest of PR is still important (guarding in the code for the subresources), but I think we'll hit the issue here the most.

coreweave/object_storage/resource_bucket_post_create_unit_test.go
```diff
@@ -0,0 +1,285 @@
...
+
+func (f *postCreateInvalidRegionFake) handleS3(w http.ResponseWriter, r *http.Request) {
+	bucket := strings.Trim(r.URL.Path, "/")
+	query := r.URL.Query()
+
+	switch {
+	case r.Method == http.MethodGet && bucket == "":
+		f.writeListBuckets(w)
+
+	case r.Method == http.MethodPut && bucket != "" && query.Has("versioning"):
```
AndresRCB, line 101: The fake only injects `InvalidRegion` on the first *Get* (readback), but every Put succeeds immediately. I think you want to fail the first `PutBucketVersioning`/`PutBucketPolicy` here too to actually test the whole thing.

coreweave/object_storage/resource_bucket_subresource_convergence_unit_test.go
```diff
@@ -0,0 +1,322 @@
...
+	if index < len(c.getErrors) && c.getErrors[index] != nil {
+		return nil, c.getErrors[index]
+	}
+	if index < len(c.getOutputs) && c.getOutputs[index] != nil {
+		return c.getOutputs[index], nil
+	}
+	return &s3.GetBucketVersioningOutput{}, nil
+}
+
+func immediateSubresourcePhaseOptions(waits *int) s3PhaseOptions {
```
AndresRCB: I like this recommendation too, so I'm sharing it verbatim:

>This is the third copy of the same helper (`immediatePhaseOptions` in resource_bucket_create_unit_test.go, `immediateS3PhaseOptions` in s3_convergence_test.go). Keep one. Same thought for the four `scripted*Client` fakes — a tiny generic `scriptedCall[In, Out]` (inputs, errors, outputs, index) would collapse ~160 lines of near-identical fake into a few lines per method.

coreweave/object_storage/resource_bucket_post_create_unit_test.go
```diff
@@ -0,0 +1,285 @@
...
+			Check: resource.ComposeAggregateTestCheckFunc(
+				resource.TestCheckResourceAttr("coreweave_object_storage_bucket.test", "name", postCreateTestBucket),
+				resource.TestCheckResourceAttr("coreweave_object_storage_bucket_versioning.test", "versioning_configuration.status", "Enabled"),
+				resource.TestCheckResourceAttr("coreweave_object_storage_bucket_policy.test", "bucket", postCreateTestBucket),
+				func(_ *terraform.State) error {
+					counts := fake.requestCounts()
+					if counts.head == 0 {
+						return fmt.Errorf("HeadBucket requests = 0, want readiness barrier before child resources")
+					}
+					if counts.getPolicy != 2 {
```
AndresRCB: `getPolicy != 2` matches here because `Check` runs before the post-apply refresh. it would be better to assert `>= 2` (or counting only pre-Check calls) because this is too specific (can fail easily later for the wrong reasons).

AndresRCB (changes requested):

coreweave/object_storage/bucket_create.go
scotchneat: Good callout. I'm going to add red tests that will cover the subresource scenarios first. Then I'll have to explore what the best course is.

coreweave/object_storage/s3_convergence.go
scotchneat: Agreed. This is what I call out in the description, which I'm looking to remediate.

coreweave/object_storage/s3_convergence.go
scotchneat: Good catch! Looking into your suggestions.

coreweave/object_storage/resource_bucket_post_create_unit_test.go
scotchneat: This one is still open. The expanded scenario requires at least five reads to cover `InvalidRegion`, an unstable match, and stable confirmation, but the assertions still require exact matches - five now.

I'm thinking this one throw while the rest will be ready for review.

coreweave/object_storage/resource_bucket_subresource_convergence_unit_test.go
scotchneat: Consolidated the timing helper and common Create/Update test harness. I kept the SDK clients typed because their request and response contracts differ. Generalizing those adapters isn’t included here.

coreweave/object_storage/resource_bucket_policy.go
scotchneat: Policy, versioning, lifecycle, and inventory now use the same convergence-client setup and options. All four have full Create and Update state-retention coverage.

coreweave/object_storage/s3_convergence.go
scotchneat: Preserved the final SDK error when the convergence budget expires. 
Persistent failures now include the phase, attempt count, SDK operation, `InvalidRegion`, and deadline.

coreweave/object_storage/bucket_create.go
scotchneat: Kept HeadBucket as the immediate minimum barrier, but not as proof that every subresource is ready. 
Each affected subresource now retries its own `InvalidRegion` and requires a stable readback before completing.

coreweave/object_storage/resource_bucket_post_create_unit_test.go
scotchneat: Updated the fake so the first policy and versioning PUTs also return `InvalidRegion` and succeed on retry. 
The test additionally covers unstable readback before convergence.

coreweave/object_storage/s3_convergence.go
scotchneat: Removed the initial delay. Mutations and the first readback run immediately. The five-second wait only occurs between a matching read and its confirmation.

coreweave/object_storage/s3_convergence.go
scotchneat: Split mutation and readback classification. Writes now fail immediately on `404` / `NoSuchBucket` / `NotFound` / `NoSuchTagSet`, while `InvalidRegion` and other transient failures remain retriable.

coreweave/object_storage/s3_convergence_test.go
```diff
@@ -0,0 +1,394 @@
...
+}
+
+func immediateS3PhaseOptions(waits *int) s3PhaseOptions {
+	return s3PhaseOptions{
+		now:   time.Now,
+		delay: func(int) time.Duration { return 0 },
+		wait: func(ctx context.Context, _ time.Duration) error {
+			*waits++
+			return ctx.Err()
+		},
```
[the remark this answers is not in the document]
scotchneat: looks like a false positive

coreweave/object_storage/s3_convergence.go
```diff
@@ -0,0 +1,324 @@
...
+		if waitingForConfirmation && delay < bucketReadbackConfirmationDelay {
+			return bucketReadbackConfirmationDelay
+		}
+		return delay
+	}
+
+	// A matching response is request-local evidence, not a service readiness
+	// guarantee. Try immediately, then require a second matching sample after
+	// a minimum confirmation delay so a match from one gateway or cache
+	// generation does not immediately end reconciliation.
```
AndresRCB: question: do we have an observed case of a *matching* readback being followed by a mismatch (match-then-drift)? The customer logs showed accepted writes followed by `InvalidRegion` on readback, which the retry already covers; the two-sample confirmation guards against a different failure. Since `reconcileBucketAfterCreate` now goes through this too, every tagged bucket create and every policy/versioning/lifecycle/inventory apply pays a fixed 5s plus one extra read. That is the same latency as the old ticker so it is not a regression, but if the drift case is hypothetical I would say so in this comment and consider a shorter `bucketReadbackConfirmationDelay` (1–2s); if it was observed, describe the observed behavior here so the constants don't get tuned away later.

coreweave/object_storage/s3_convergence.go
```diff
@@ -0,0 +1,324 @@
...
+		options.wait = waitForS3Backoff
+	}
+	return options
+}
+
+type s3PhaseMetadata struct {
+	phase                   string
+	bucket                  string
+	zone                    string
+	sharedPropagationBudget bool
```
AndresRCB: `sharedPropagationBudget` is only set in `runBucketReadbackPhase` (line 165) and overwrites whatever callers passed with `true` (the other uses/locations never use this value as it was set). It doesn't seem like you're using it further downstream when other parts of the code instantiate or need `s3PhaseMetadata`. I think you can just make it a local argument in the code using sharedPropagationBudget and remove it from here.

It just seems like it's not really being used. This is a nit, btw (the code works fine as-is)

coreweave/object_storage/resource_bucket_inventory.go
```diff
@@ -45,6 +44,11 @@ const (
 	ErrNoSuchInventoryConfiguration string = "NoSuchInventoryConfiguration"
 )
 
+type bucketInventoryConfigurationClient interface {
```
AndresRCB: This is a nit but it was a bit confusing: I see 4 similar places with similar interfaces and fixes:
- `bucketPolicyAPI`
- `bucketVersioningAPI`
- `bucketInventoryConfigurationClient`
- `bucketLifecycleConfigurationClient`

The first two use `"... application"` in their messages, while the last two use `"... write"` in their messages. I would prefer if they all used the same naming conventions (either they're APIs or ConfigurationClients, and they all either "application" or "write").

The interface names are less important than the `"... write"` messages, though, because those last ones are for users.

AndresRCB (approved):

PR #436, opened 2026-09-01
feat(workload-federation): add OIDC config data source

hghafghazi:
## Summary

Implements the workload federation OIDC configuration data source for [CLDIAM-598](https://coreweave.atlassian.net/browse/CLDIAM-598).

- Adds the `coreweave_workload_federation_oidc_config` data source.
- Supports lookup by either:
  - Stable `id`.
  - The unique `issuer_url` and `audience` combination.
- Scopes issuer/audience lookup to the organization associated with the provider’s authenticated context.
- Requires exactly one selector: `id`, or both `issuer_url` and `audience`.
- Exposes OIDC configuration metadata, including organization, name, description, and activation state.
- Shares the model with the existing OIDC configuration resource without introducing a redundant public `uid` attribute.
- Provides clear diagnostics for missing or ambiguous lookup results.
- Adds examples and generated provider documentation.

The service-account and trust-configuration data sources remain dependent on their corresponding draft provider and backend work.

## Testing

Added Terraform protocol-level coverage for:

- ID lookup.
- Issuer URL and audience lookup.
- Missing lookup results.
- Missing, incomplete, and conflicting selectors.
- Empty, invalid, and oversized selector values.
- Deactivated configurations.
- Defensive handling of duplicate issuer/audience results.

Validation performed:

- `go test ./...`
- `golangci-lint run`
- `git diff --check`

## References

- [CLDIAM-598](https://coreweave.atlassian.net/browse/CLDIAM-598)

[CLDIAM-598]: https://coreweave.atlassian.net/browse/CLDIAM-598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


coreweave/workload_federation/resource_oidc_config.go
```diff
@@ -103,6 +104,13 @@ func (r *OIDCConfigResource) Schema(_ context.Context, _ resource.SchemaRequest,
 					stringplanmodifier.UseStateForUnknown(),
 				},
 			},
+			"uid": schema.StringAttribute{
```
AndresRCB: You're adding a public `uid` that is identical to `id` only so the model struct can be shared. `coreweave_cks_cluster` / `coreweave_networking_vpc` data sources take `id` as the lookup key and make `id` Optional+Computed on the data source. You can drop `uid` from both schemas to keep the shared model without adding these redundant fields.

internal/provider/provider.go
```diff
@@ -256,6 +256,7 @@ func (p *CoreweaveProvider) DataSources(ctx context.Context) []func() datasource
 		inference.NewInferenceDeploymentParametersDataSource,
 		inference.NewCapacityClaimParametersDataSource,
 		inference.NewGatewayParametersDataSource,
+		workloadfederation.NewOIDCConfigDataSource,
```
AndresRCB, line 259: A new data source is a feature, so the PR title should be `feat(workload-federation): ...` rather than `fix(...)` (this is important so release-please publishes a `minor` and not a `patch`)

AndresRCB (changes requested):

blairdrummond (approved):

AndresRCB (approved):

PR #440, opened 2026-09-09
feat(sandbox): add v1 managed runner resource

brandonrjacobs:
The provider currently cannot manage Sandbox runners. Add `coreweave_sandbox_managed_runner` using the v1 managed runner endpoint, with the runner's policy required on the same resource.

- Expose typed runner identity, release channel, maintenance, deployment overrides/scaling/runtime classes, resource-limit enforcement, volumes, tenant metrics, image pull behavior, and all direct endpoint modes.
- Expose the full v1 policy: runtime base, resource/image/network/security/instance/lifecycle/metadata/volume constraints, and control-plane source IP restrictions. Omitted GPU caps and empty source allowlists retain their distinct meanings.
- Support import, immutable identity replacement, drift detection, explicit field masks for updates and clearing settings, policy etags, idempotent create requests, and waiting for asynchronous deletion with a reusable polling ticker. Read-only deployment and endpoint diagnostics are available in state.
- Match Sandman's runner-group default during planning: omitted or null `runner_group_id` resolves to `default`, removing a custom group resets it to `default`, and an explicit empty string is rejected before applying. This prevents inconsistent Terraform state against the deployed server.
- Require lowercase `zone` values before creating a runner so the API response matches Terraform's planned value. Use `lower(...)` for CKS zone references; the generated example normalizes its zone input.
- Reuse provider authentication, endpoint configuration, and retry handling. Include generated documentation, usage/import examples, schema coverage checks against the pinned v1 descriptors, and Terraform CLI lifecycle tests against a local Connect server.

Legacy profile bindings, templates, configuration flags, and version outputs are excluded. Policy remains embedded on the runner; there is no independently managed policy resource.

**API availability:** [aviato#2084](https://github.com/coreweave/aviato/pull/2084) merged on September 9 and is included in Sandman `v0.177.0`. [sandbox-deploy#756](https://github.com/coreweave/sandbox-deploy/pull/756) promoted that version to production and staging. On September 10, regional observability confirmed healthy Sandman containers on `v0.177.0` in both environments and successful migration jobs using the same release. The deployment dependency is satisfied.

Validation:

- `make build`, `make test`, and `make lint` pass on the updated provider. Repository-wide tests include real Terraform create/read/update/import/replacement/destroy against a local Connect server, endpoint modes, drift, remote deletion, unknown inputs, and stale-etag failures.
- The local API fixture now reproduces deployed runner-group normalization. A regression test failed before the fix with `Provider produced inconsistent result after apply` when the group was omitted. Coverage now verifies default-group creation, a no-change plan, import, custom-group updates, resetting a custom group, null configuration, and validation of empty strings before any create request.
- Terraform CLI tests explicitly verify that a whole policy or nested constraint can be unknown during planning: values resolve before create/update, unchanged values send no mutation, changed values send exactly one policy update with the existing etag check, and the following plan is empty. Configured policy expressions remain unknown until resolved instead of being replaced by previous state.
- The API fixture now also reproduces Aviato v1 zone normalization. The uppercase regression failed before the fix with an inconsistent-result error after creation. Uppercase, mixed-case, and initially unknown uppercase zones are now rejected without creating a runner; `lower(...)` on an initially unknown CKS-style zone succeeds through creation, empty plans, and import.
- `make generate` passed and regenerated the runner resource documentation. `go mod tidy -diff` and `git diff --check` are clean.
- Earlier backend compatibility validation for aviato#2084 passed the provider's full runner fixture through create, combined policy/configuration updates, no-change plans, and destroy against the actual Sandman handler and local CockroachDB, with cluster resolution/preflight simulated.

Live validation on September 10 used Terraform 1.15.9 and the local provider build at `15d79c1` with the `cw9704` test organization's API key against `https://api.coreweave.com`:

- Imported the existing `test-runner-lab-05` runner into isolated local state, generated configuration from its imported mutable attributes, and verified an empty baseline plan.
- Verified that the saved plan contained exactly one in-place change, to `display_name`, then applied it successfully. A subsequent plan was empty.
- Applied a second display-name-only plan to restore the original value and verified a final empty plan. Policy, spec, runner identity/group, policy etag, and deployment revisions matched the original values.
- Removed the resource from the isolated local state after restoration. No runner was created, deleted, upgraded, or redeployed during this check.

The prior CI run passed build, unit tests, lint, generation, and snapshot-build checks. Its acceptance workflow was blocked by the unrelated CKS VPC sweep failure. Live Sandbox creation and deletion have not been exercised.


coreweave/sandbox/runner_schema.go
```diff
@@ -0,0 +1,272 @@
...
+		MarkdownDescription: "Policy governing every sandbox on this runner. Required even for the explicit empty posture `policy = {}`. Policy is updated with an etag to detect concurrent changes.",
+		Attributes: map[string]schema.Attribute{
+			"display_name": optionalString("Human-readable policy name."),
+			"base":         base,
+			"constraints":  policyConstraintsAttribute(),
+			"control_plane_access": optionalObject("Request-time controls in addition to authentication and tenant authorization.", map[string]schema.Attribute{
+				"source_ip_allowlist": optionalObject("Public source IP restriction. A present empty object denies every public source; an absent object imposes no source IP restriction.", map[string]schema.Attribute{"cidrs": cidrs}),
+			}),
+		},
+	}
```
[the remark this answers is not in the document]
brandonrjacobs: Verified this behavior with Terraform CLI tests in 15d79c1 and retained the configured policy semantics. ValidateConfig already defers conversion while any policy value is unknown; Terraform resolves those expressions before Create/Update. The pinned framework's UseStateForUnknown implementation explicitly skips unknown configuration, and a custom modifier copying old policy state could conceal a real policy change. See [HashiCorp's plan-modification guidance](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#common-use-case-attribute-plan-modifiers).

Expanded TestManagedRunnerUnknownPolicyResolvesBeforeApply to cover both the whole policy and a nested resource constraint supplied by terraform_data.output. Each test asserts the value is unknown in the plan, then verifies successful creation, no API mutation when replacement resolves to the same policy, exactly one policy update with the existing etag check when it resolves differently, and a subsequent empty plan. Both cases pass, along with make build, make test, and make lint. Added a schema comment explaining why this input does not use prior-state substitution.

coreweave/sandbox/resource_managed_runner.go
```diff
@@ -0,0 +1,415 @@
...
+		}
+		timer := time.NewTimer(2 * time.Second)
+		select {
+		case <-ctx.Done():
+			timer.Stop()
+			resp.Diagnostics.AddError("Managed Runner Deletion Incomplete", ctx.Err().Error())
+			return
+		case <-timer.C:
+		}
+	}
```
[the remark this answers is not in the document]
brandonrjacobs: Fixed in 15d79c1: deletion now allocates one time.Ticker outside the loop and defers Stop, retaining the immediate initial lookup, context cancellation, and 30-minute timeout. The Terraform lifecycle test still verifies asynchronous deletion completes after two polls. Focused tests, make build, make test, and make lint pass.

coreweave/sandbox/resource_managed_runner.go
```diff
@@ -0,0 +1,415 @@
...
+		return fmt.Errorf("spec: %w", err)
+	}
+	m.Policy, err = objectFromProto(ctx, runner.Policy, m.Policy, policyAttribute().GetType().(types.ObjectType))
+	if err != nil {
+		return fmt.Errorf("policy: %w", err)
+	}
+	m.ID = types.StringValue(runner.Identity.RunnerId)
+	m.RunnerID = m.ID
+	m.RunnerGroupID = optionalResponseString(m.RunnerGroupID, runner.Identity.RunnerGroupId)
+	m.Zone = types.StringValue(runner.Identity.Zone)
```
scott-cw, line 239: **High: Uppercase zones fail after successful creation**

`zone = "US-EAST-04A"` passes validation, but the [backend lowercases it during creation](https://github.com/coreweave/aviato/blob/main/internal/sandman/v1_runner_handler.go#L74). This assignment writes the lowercase response into state, changing Terraform's known planned value after the runner has been created.

An isolated Terraform regression test with the API fixture reproducing that normalization failed with:

```text
Provider produced inconsistent result after apply
.zone: was cty.StringVal("US-EAST-04A"), but now cty.StringVal("us-east-04a")
```

Reject noncanonical zones before creation, or preserve case-equivalent configured values during readback. Add regression coverage with the server's actual normalization behavior.

scott-cw (commented):
Runner creation has a state-consistency issue when the server normalizes a configured zone.

coreweave/sandbox/resource_managed_runner.go
```diff
@@ -0,0 +1,415 @@
...
+			}
+			mask.Paths = append(mask.Paths, "policy")
+			runner.Etag = previous.Etag.ValueString()
+		}
+	}
+	if !m.Spec.IsUnknown() && !m.Spec.Equal(previous.Spec) {
+		for name, value := range m.Spec.Attributes() {
+			old, ok := previous.Spec.Attributes()[name]
+			if !value.IsUnknown() && (!ok || !value.Equal(old)) {
+				mask.Paths = append(mask.Paths, "spec."+name)
```
AndresRCB: consider: suppress spec-mask entries when the old and new values encode to the same protobuf value, as the policy path already does. Removing `enforce_resource_limits = false` sends `spec.enforce_resource_limits` even though both requests encode the same `false`; I reproduced this with the Terraform CLI fixture. The backend treats this mask as a deployment update and creates/advances a runner revision, so removing a redundant setting causes an unnecessary deployment revision. Compare each changed top-level spec field using protobuf semantics, preserving message presence, and add a regression asserting that this transition sends no update.

AndresRCB (dismissed):

coreweave/sandbox/resource_managed_runner.go
brandonrjacobs: Fixed in 67dda4a. Confirmed Aviato v1 lowercases zones on creation and returns the stored lowercase value; cwic's v1 interactive CKS selection also lowercases the zone. The legacy v1beta2 response formatter uses uppercase, but this resource uses v1.

The provider now rejects noncanonical casing with an attribute error before creating a runner and directs users to `lower(...)` for CKS zone references. Updated the generated documentation and example accordingly. The API fixture now performs the same lowercase normalization as Aviato.

Reproduced the reported inconsistent-result error before the fix. Terraform CLI regression tests now verify uppercase, mixed-case, and initially unknown uppercase values are rejected with zero runner creations. An initially unknown value wrapped in `lower(...)` successfully creates, imports, and produces empty subsequent plans. `make build`, `make test`, `make lint`, `make generate`, `go mod tidy -diff`, and `git diff --check` all pass locally.

Sources: [v1 normalization](https://github.com/coreweave/aviato/blob/v0.179.0/internal/sandman/v1_runner_handler.go#L74), [v1 response](https://github.com/coreweave/aviato/blob/v0.179.0/internal/sandman/v1_runner_convert.go#L47), [cwic interactive form](https://github.com/coreweave/cwic-internal/blob/ac2eefb3b9349f9415a566d7e6935f1c1b1e3675/cmd/sandbox/runner_create.go#L216).

AndresRCB (approved):

PR #443, opened 2026-09-09
feat(object-storage): manage individual access keys

AndresRCB:
Adds `coreweave_object_storage_access_key` for creating, importing, refreshing, and individually revoking Object Storage credentials. Duration and attributes are creation settings; changes replace the key. The creation-only secret survives refresh, while imports leave the secret and original duration unavailable. Expired and inactive keys remain managed for explicit rotation; missing keys and retained `DELETED` records are removed from state.

Uses the existing CWObject client, disables retries for credential creation, and preserves the new ID and secret if a follow-up read fails. Includes validation, lifecycle and import tests, examples, and generated documentation. Disables acceptance-matrix fail-fast so one suite's cleanup failure does not cancel independent suites; each suite still requires its own cleanup to pass.

Implements [CLDAPI-1115](https://coreweave.atlassian.net/browse/CLDAPI-1115).

Validation: repository unit tests (including Terraform CLI against a local fake API), pinned lint, build, generation, and dependency tidiness pass. Focused live `TestAccAccessKeyResource` passes in internal org `cw0000`, covering permanent and expiring keys, stable plans, duration/attribute replacement, import, and individual revocation. The unrelated CKS acceptance cleanup remains blocked by an allocated QA VPC prefix.

[CLDAPI-1115]: https://coreweave.atlassian.net/browse/CLDAPI-1115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


msillz (approved):
approved.

PR #446, opened 2026-09-11
feat(object_storage): add bucket capacity caps support to Terraform

zzhao-cloud:
## Overview

Adds `capacity_cap_bytes` to object storage bucket settings in Terraform.

Create a 1 TiB cap:

```hcl
resource "coreweave_object_storage_bucket_settings" "example" {
  bucket             = coreweave_object_storage_bucket.example.name
  capacity_cap_bytes = 1099511627776
}
```

To update it, change the number and run `terraform apply` again.

These semantics provide a safe, predictable customer experience:

- An omitted cap value is ignored, so an existing cap stays.
- An out-of-band (created out of terraform) cap is observed but not changed when omitted; it will be updated if hcl has a different value.
- Destroying bucket settings (not bucket) via terraform keeps the cap.
- Deleting the bucket removes the cap without requiring cap permission (true for both API and Trraform).
- The organization must be entitled to create or update caps.

### DIFF
--- a/coreweave/object_storage/resource_bucket_settings.go
+++ b/coreweave/object_storage/resource_bucket_settings.go
@@ -11,6 +11,7 @@ import (
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/hashicorp/hcl/v2/hclwrite"
 	"github.com/hashicorp/terraform-plugin-framework-validators/int32validator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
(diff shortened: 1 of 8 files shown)


coreweave/object_storage/resource_bucket_settings.go
```diff
@@ -86,9 +95,22 @@ func (s *BucketSettingsModel) ToProtoObject() *cwobjectv1.CWObjectBucketSettings
...
+	if !s.CapacityCapBytes.IsNull() && !s.CapacityCapBytes.IsUnknown() {
+		settings.SetCapacityCapBytes(uint64(s.CapacityCapBytes.ValueInt64())) //nolint:gosec
+	}
 	return &settings
 }
 
+// An omitted cap can still appear in the plan from state, including an
+// out-of-band value. Exclude it so unrelated updates neither change the cap
+// nor require capacity-cap permission.
+func omitUnconfiguredCapacityCap(settings *cwobjectv1.CWObjectBucketSettings, configured types.Int64) {
```
zzhao-cloud, line 108: this is to avoid this blocking ux:

user A create a bucket via hcl -> user B then apply bucket cap via API (out-of-band cap) -> user A try to use the (drifted) hcl to edit auditLogging -> Terraform refresh state and see the out-of-band cap -> now user A find that bucket cap permission is required to update an unrelated setting.

PR #447, opened 2026-09-11
feat(object_storage): add bucket capacity caps support to Terraform

zzhao-cloud:
## Overview

Adds `capacity_cap_bytes` to object storage bucket settings in Terraform.

Create a 1 TiB cap:

```hcl
resource "coreweave_object_storage_bucket_settings" "example" {
  bucket             = coreweave_object_storage_bucket.example.name
  capacity_cap_bytes = 1099511627776
}
```

To update it, change the number and run `terraform apply` again.

These semantics provide a safe, predictable customer experience:

- An omitted cap value is ignored, so an existing cap stays.
- An out-of-band (created out of terraform) cap is observed but not changed when omitted; it will be updated if hcl has a different value.
- Destroying bucket settings (not bucket) via terraform keeps the cap.
- Deleting the bucket removes the cap without requiring cap permission (true for both API and Terraform).
- The organization must be entitled to create or update caps.

## Test
Verified locally hitting staging and Prod CAIOS APIs, cases covered for create, update, destroy, omission behavior and out-of-band cap handling behaviors

### DIFF
--- a/coreweave/object_storage/resource_bucket_settings.go
+++ b/coreweave/object_storage/resource_bucket_settings.go
@@ -11,6 +11,7 @@ import (
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/hashicorp/hcl/v2/hclwrite"
 	"github.com/hashicorp/terraform-plugin-framework-validators/int32validator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
@@ -43,6 +44,7 @@ type BucketSettingsModel struct {
 	AuditLoggingEnabled        types.Bool   `tfsdk:"audit_logging_enabled"`
 	ArchiveEnabled             types.Bool   `tfsdk:"archive_enabled"`
 	ArchiveAfterLastAccessDays types.Int32  `tfsdk:"archive_after_last_access_days"`
+	CapacityCapBytes           types.Int64  `tfsdk:"capacity_cap_bytes"`
 }
 
 // BucketSettingsResourceModel is an alias for BucketSettingsModel for consistency with other resources
@@ -70,6 +72,13 @@ func (s *BucketSettingsModel) Set(settings *cwobjectv1.CWObjectBucketSettings) {
 	} else {
 		s.ArchiveAfterLastAccessDays = types.Int32Null()
 	}
+
+	// Read the cap from the read-only configured field (the write oneof isn't returned).
+	if settings.ConfiguredCapacityCapBytes != nil {
+		s.CapacityCapBytes = types.Int64Value(int64(settings.ConfiguredCapacityCapBytes.Value)) //nolint:gosec
+	} else {
+		s.CapacityCapBytes = types.Int64Null()
+	}
 }
 
 func (s *BucketSettingsModel) ToProtoObject() *cwobjectv1.CWObjectBucketSettings {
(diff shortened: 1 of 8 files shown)


coreweave/object_storage/resource_bucket_settings.go
```diff
@@ -86,9 +95,22 @@ func (s *BucketSettingsModel) ToProtoObject() *cwobjectv1.CWObjectBucketSettings
...
+	if !s.CapacityCapBytes.IsNull() && !s.CapacityCapBytes.IsUnknown() {
+		settings.SetCapacityCapBytes(uint64(s.CapacityCapBytes.ValueInt64())) //nolint:gosec
+	}
 	return &settings
 }
 
+// An omitted cap can still appear in the plan from state, including an
+// out-of-band value. Exclude it so unrelated updates neither change the cap
+// nor require capacity-cap permission.
+func omitUnconfiguredCapacityCap(settings *cwobjectv1.CWObjectBucketSettings, configured types.Int64) {
```
zzhao-cloud, line 109: this is to avoid this blocking ux:

user A create a bucket via hcl -> user B then apply bucket cap via API (out-of-band cap) -> user A try to use the (drifted) hcl to edit auditLogging -> Terraform refresh state and see the out-of-band cap -> now user A find that bucket cap permission is required to update an unrelated setting.

coreweave/object_storage/resource_bucket_settings.go
```diff
@@ -240,6 +264,15 @@ func (b *BucketSettingsResource) Schema(ctx context.Context, req resource.Schema
 					int32validator.AtLeast(1),
 				},
 			},
+			// Omission leaves the cap unchanged.
+			"capacity_cap_bytes": schema.Int64Attribute{
+				MarkdownDescription: "Maximum number of STANDARD-class bytes the bucket may store. New STANDARD writes are rejected once bucket usage would exceed this cap; `0` is a valid cap that blocks all new STANDARD writes. Omit to leave the cap unchanged. Removing this resource leaves any cap in place; the cap is cleared when the bucket itself is deleted. Your organization must be entitled to configure this setting.",
+				Optional:            true,
+				Computed:            true,
```
AndresRCB, line 272: Do you want to use `int64planmodifier.UseStateForUnknown()` here? Without it, changes to other fields will show `capacity_cap_bytes = (known after apply)` instead of using what the state has. Not necessary, though.

```suggestion
				Computed:            true,
				PlanModifiers: []planmodifier.Int64{
					int64planmodifier.UseStateForUnknown(),
				},
```

coreweave/object_storage/resource_bucket_settings_unit_test.go
```diff
@@ -619,3 +627,198 @@ func TestBucketSettingsArchiveWithoutRetentionRejectedAtValidate(t *testing.T) {
...
+		capacityCapBytes: int64Ptr(-1),
+	})
+
+	require.True(t, diagsHaveErrors(diags), "a negative cap must be rejected")
+	assert.Contains(t, diagText(diags), "at least 0")
+}
+
+// TestBucketSettingsOmittedCapacityCapIsNotResent verifies the wire-level no-op:
+// Optional+Computed preserves a remotely observed cap in the plan, so an
+// unrelated update must consult configuration before writing.
```
AndresRCB: nit: this comment doesn't match what the framework does today. Without a plan modifier the observed cap is not preserved in the plan; it is replaced with unknown (see my comment on the schema). Worth asserting the planned `capacity_cap_bytes` value in this test so it pins whichever behavior you land on, rather than only checking the wire request.

coreweave/object_storage/resource_bucket_settings_test.go
```diff
@@ -56,6 +56,7 @@ func createBucketSettingsTestStep(ctx context.Context, t *testing.T, opts bucket
 		AuditLoggingEnabled:        opts.settings.AuditLoggingEnabled,
 		ArchiveEnabled:             opts.settings.ArchiveEnabled,
 		ArchiveAfterLastAccessDays: opts.settings.ArchiveAfterLastAccessDays,
+		CapacityCapBytes:           opts.settings.CapacityCapBytes,
```
AndresRCB, line 59: `TestBucketSettingsResource` doesn't set `capacity_cap_bytes` in any step despite this setup here, so the API calls don't test this new feature (but it tests that omitting it doesn't break anything). You may want to update the steps below to try at least one call with `capacity_cap_bytes`.

coreweave/object_storage/resource_bucket_settings.go
```diff
@@ -240,6 +264,15 @@ func (b *BucketSettingsResource) Schema(ctx context.Context, req resource.Schema
 					int32validator.AtLeast(1),
 				},
 			},
+			// Omission leaves the cap unchanged.
+			"capacity_cap_bytes": schema.Int64Attribute{
+				MarkdownDescription: "Maximum number of STANDARD-class bytes the bucket may store. New STANDARD writes are rejected once bucket usage would exceed this cap; `0` is a valid cap that blocks all new STANDARD writes. Omit to leave the cap unchanged. Removing this resource leaves any cap in place; the cap is cleared when the bucket itself is deleted. Your organization must be entitled to configure this setting.",
```
AndresRCB, line 270: If `0` blocks all writes and removing the field keeps the existing cap, how can users clear the cap with TF?

zzhao-cloud: there is no way to clear cap with TF, this is a trade-off we did after careful consideration. 

The main concern is user would accidentally remove a (out-of-band) cap with a outdated template. To avoid that, we remove the capability of removing cap via TF to avoid confusion of omission/null/blank cap etc.. 

We know that remove a cap won't be the main use case of this feature.,and user can still use API to remove a cap if needed. For 0 cap, user could just update the value to a higher one.

coreweave/object_storage/resource_bucket_settings_test.go
zzhao-cloud: good catch, thanks. I updated test to cover this feature

coreweave/object_storage/resource_bucket_settings.go
zzhao-cloud: good catch, fixed.

AndresRCB (approved):

PR #449, opened 2026-09-15
feat(inference): Wait for the STATUS_UPDATING rollout to complete on update

kulkarnisamr:
Fixes to incorporate `STATUS_UPDATING`:

- Update previously returned once resources were applied. Now that the API surfaces
STATUS_UPDATING during a revision rollout (and it outranks STATUS_READY until the new
revision serves), hold Update through the rollout and complete on STATUS_READY so an
apply reflects the rolled-out version. A first-poll grace lets a template/engine change
surface STATUS_UPDATING before sampling, avoiding completion on the pre-rollout READY;
scale-only changes stay READY (template_only semantics) and complete right after it.

- Update now blocks on the full STATUS_UPDATING rollout, which for large models can
exceed the prior 20m budget; match Create's 45m so a slow revision rollout does not
time out the apply.

- An update that sets disabled=true tears the deployment down, so its status
never reaches STATUS_READY (it settles at STATUS_CREATING). The Update waiter
targeting STATUS_READY therefore polled until the 45m timeout, deadlocking the
acceptance suite (TestInferenceDeployment/lifecycle). Gate the waiter on the
desired state: a disabled deployment completes on ResourcesApplied like Create,
while a serving deployment still holds through the rollout to STATUS_READY.


coreweave/inference/resource_inference_deployment.go
```diff
@@ -80,9 +80,26 @@ var (
...
+	// so deletion can be polled deterministically via status rather than CodeNotFound.
+	deletedState = "NOT_FOUND"
+
+	// rolloutStartGracePeriod delays Update's first status poll so a template or engine
+	// change has time to surface STATUS_UPDATING before we sample it. The worker picks up
+	// a spec change on its broker delta poll (5s) and then flips to STATUS_UPDATING;
+	// without this grace Update could complete on the pre-rollout STATUS_READY still
+	// reported against the prior revision. A scale-only change never leaves READY
+	// (template_only semantics) and completes right after this delay.
+	rolloutStartGracePeriod = 30 * time.Second
```
AndresRCB: blocking: A fixed 30s delay still allows Update to report success on the previous revision's READY status if the worker takes longer to process the new spec (for example during backlog or restart); the 5s poll interval is not an upper bound on processing latency. With a stub that stays READY until 35s, then goes UPDATING until 40s, the actual waiter returns success after its first poll at 30s, before the rollout even starts. Please gate completion on an acknowledgement tied to the requested revision/spec (which may require API support), while retaining the scale-only path, and add a delayed-worker regression test. Increasing the grace period would only move the race.

AndresRCB (approved):

coreweave/inference/resource_inference_deployment.go
kulkarnisamr: right now a spec change stands up a whole new revision-scoped DGD and the old one keeps serving (`Ready=True`) while traffic weights shift over to the new revision. Server-side, the `InternalUpdating` marker forces the derived status to `UPDATING` and holds it there until the new revision is actually "servable" and routes have flipped - so we can't complete mid-shift on the old revision's `READY`. The only vulnerable moment is this initial pre-UPDATING gap until the broker picks the delta up.

agreed, the right fix would be to have the API acknowledge the requested revision (an observedGeneration/revision field on status), then poll for observed >= requested && READY, and it also cleanly handles the scale-only case. we have a few more changes going on with rolling updates so I'll file a follow-up to add this and switch the waiter over once it lands.

